PR #46, opened 2017-12-12
Trying a few things

etpinard:
... mostly from https://github.com/electron/electron/issues/1277#issuecomment-296397800 but to no avail. @scjody 's diagnostic from https://github.com/plotly/image-exporter/issues/39#issue-280597766 is prevails.

---

Note that I tried docker building a few [versions](https://github.com/electron/electron/releases) of electron including some pretty older versions (with much older chromium like nw.js is using :wink: ), but to no avail also. This PR locks `electron` to its latest non-beta release.

### DIFF
--- a/bin/plotly-export-server_electron.js
+++ b/bin/plotly-export-server_electron.js
@@ -94,6 +94,15 @@ app.on('export-error', (info) => {
   }
 })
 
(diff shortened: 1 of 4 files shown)


etpinard:
Taken over by https://github.com/plotly/image-exporter/pull/78

PR #55, opened 2018-02-13
Catch hanging figure payloads in parse step

etpinard:
A proof of concept for https://github.com/plotly/streambed/issues/9865#issuecomment-365322166 for more info.

cc @scjody @alexcjohnson

### DIFF
--- a/src/component/plotly-graph/parse.js
+++ b/src/component/plotly-graph/parse.js
@@ -114,6 +114,12 @@ function parseDim (result, opts, dim) {
 
 function willFigureHang (result) {
   const data = result.figure.data
+
+  // cap the number of traces
+  if (data.length > 200) {
+    return true
+  }
+
   const numberOfPtsPerTraceType = {}
 
   for (var i = 0; i < data.length; i++) {
--- a/src/component/plotly-graph/parse.js
+++ b/src/component/plotly-graph/parse.js
@@ -122,7 +122,7 @@ function willFigureHang (result) {
 
   const numberOfPtsPerTraceType = {}
 
-  for (var i = 0; i < data.length; i++) {
+  for (let i = 0; i < data.length; i++) {
     const trace = data[i] || {}
     const type = trace.type || 'scatter'
 
@@ -130,25 +130,42 @@ function willFigureHang (result) {
       numberOfPtsPerTraceType[type] = 0
     }
 
-    const lengths = Object.keys(trace)
-      .filter(k => Array.isArray(trace[k]))
-      .map(k => trace[k].length)
+    numberOfPtsPerTraceType[type] += estimateDataLength(trace)
 
-    // Consider the array of maximum length as a proxy to determine
-    // the number of points to be drawn. In general, this estimate
-    // can be (much) smaller than the true number of points plotted
-    // when it does not match the length of the other coordinate arrays.
-    numberOfPtsPerTraceType[type] += Math.max(...lengths)
-
+    // cap the number of point per trace type
     if (numberOfPtsPerTraceType[type] > maxPtsPerTraceType(type)) {
       return true
     }
   }
 }
 
+// Consider the array of maximum length as a proxy to determine
+// the number of points to be drawn. In general, this estimate
+// can be (much) smaller than the true number of points plotted
+// when it does not match the length of the other coordinate arrays.
+function findMaxArrayLength (cont) {
+  const lengths = Object.keys(cont)
+    .filter(k => Array.isArray(cont[k]))
+    .map(k => cont[k].length)
+
+  return Math.max(...lengths)
+}
+
+function estimateDataLength (trace) {
+  // special case for e.g. parcoords and splom traces
+  if (Array.isArray(trace.dimensions)) {
+    return trace.dimensions
+      .map(findMaxArrayLength)
+      .reduce((a, v) => a + v)
+  }
+
+  return findMaxArrayLength(trace)
+}
+
 function maxPtsPerTraceType (type) {
   switch (type) {
     case 'scattergl':
+    case 'splom':
     case 'pointcloud':
       return 1e7
 
--- a/src/component/plotly-graph/parse.js
+++ b/src/component/plotly-graph/parse.js
@@ -136,6 +136,17 @@ function willFigureHang (result) {
     if (numberOfPtsPerTraceType[type] > maxPtsPerTraceType(type)) {
       return true
     }
+
+    // other special cases
+
+    // mesh3d will `alphahull` and 1000+ pts
+    if (
+      type === 'mesh3d' &&
+      'alphahull' in trace && Number(trace.alphahull) >= 0 &&
+      findMaxArrayLength(trace) > 1000
+    ) {
+      return true
+    }
   }
 }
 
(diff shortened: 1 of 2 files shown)


src/component/plotly-graph/parse.js
```diff
@@ -108,4 +112,99 @@ function parseDim (result, opts, dim) {
...
+      .reduce((a, v) => a + v)
+  }
+
+  return findMaxArrayLength(trace)
+}
+
+function maxPtsPerTraceType (type) {
+  switch (type) {
+    case 'scattergl':
+    case 'splom':
```
alexcjohnson: Not that we really have this info yet, but I'd have thought `splom` would need to be limited quite a bit lower, perhaps the same as `parcoords` provisionally but potentially even lower.

src/component/plotly-graph/parse.js
```diff
@@ -108,4 +112,99 @@ function parseDim (result, opts, dim) {
...
+}
+
+// Consider the array of maximum length as a proxy to determine
+// the number of points to be drawn. In general, this estimate
+// can be (much) smaller than the true number of points plotted
+// when it does not match the length of the other coordinate arrays.
+function findMaxArrayLength (cont) {
+  const lengths = Object.keys(cont)
+    .filter(k => Array.isArray(cont[k]))
+    .map(k => cont[k].length)
```
alexcjohnson: 2D arrays?

src/component/plotly-graph/parse.js
```diff
@@ -108,4 +112,99 @@ function parseDim (result, opts, dim) {
...
+      return 5e5
+    case 'scattermapbox':
+      return 5e5
+
+    case 'histogram':
+    case 'histogram2d':
+    case 'histogram2dcontour':
+    case 'box':
+    case 'violin':
+      return 1e5
```
alexcjohnson: Can these really only handle 1e5? They're not actually displaying every point, I'd have thought we could bump up to 1e6 perhaps.

src/component/plotly-graph/parse.js
```diff
@@ -108,4 +112,99 @@ function parseDim (result, opts, dim) {
...
+
+  return Math.max(...lengths)
+}
+
+function estimateDataLength (trace) {
+  // special case for e.g. parcoords and splom traces
+  if (Array.isArray(trace.dimensions)) {
+    return trace.dimensions
+      .map(findMaxArrayLength)
+      .reduce((a, v) => a + v)
```
alexcjohnson: Nicely done

src/component/plotly-graph/parse.js
```diff
@@ -108,4 +112,99 @@ function parseDim (result, opts, dim) {
...
+    if (!(type in numberOfPtsPerTraceType)) {
+      numberOfPtsPerTraceType[type] = 0
+    }
+
+    numberOfPtsPerTraceType[type] += estimateDataLength(trace)
+
+    // cap the number of point per trace type
+    if (numberOfPtsPerTraceType[type] > maxPtsPerTraceType(type)) {
+      return true
+    }
```
alexcjohnson: I'm imagining someone *trying* to hang our servers, sending in one trace of each type, each just under our limit... do we want to do something like:
```js
budget += estimateDataLength(trace) / maxPtsPerTraceType(type);
if(budget > 1) return true;
```

src/component/plotly-graph/parse.js
```diff
@@ -108,4 +112,99 @@ function parseDim (result, opts, dim) {
...
+// Consider the array of maximum length as a proxy to determine
+// the number of points to be drawn. In general, this estimate
+// can be (much) smaller than the true number of points plotted
+// when it does not match the length of the other coordinate arrays.
+function findMaxArrayLength (cont) {
+  const lengths = Object.keys(cont)
+    .filter(k => Array.isArray(cont[k]))
+    .map(k => cont[k].length)
+
+  return Math.max(...lengths)
```
alexcjohnson: Better fall back to 0 if there are no arrays - otherwise the user could start with an empty trace (`Math.max() = -Infinity`) and every trace after that would be ignored.

src/component/plotly-graph/parse.js
etpinard: Oh right. We can get a better estimate for those. Thanks for noticing :+1:

src/component/plotly-graph/parse.js
etpinard: I was thinking of `box` and `violin` for (box)points turned on. We could make a special case for those I guess.

alexcjohnson: Ah good call - most of the time outliers are a tiny fraction of total points, but not always...

etpinard:
@alexcjohnson thanks for the feedback! I think all your comments have been addressed in 884afdf

src/component/plotly-graph/parse.js
```diff
@@ -155,11 +167,23 @@ function willFigureHang (result) {
...
 function findMaxArrayLength (cont) {
-  const lengths = Object.keys(cont)
+  const arrays = Object.keys(cont)
     .filter(k => Array.isArray(cont[k]))
-    .map(k => cont[k].length)
+    .map(k => cont[k])
 
-  return Math.max(...lengths)
+  const lengths = arrays.map(arr => {
+    const innerArrays = arr.filter(Array.isArray)
```
alexcjohnson: isn't this going to be super slow for big 1D arrays? Most if not all of our modules that support 1D or 2D differentiate just based on the first element.

src/component/plotly-graph/parse.js
```diff
@@ -120,30 +120,42 @@ function willFigureHang (result) {
...
-    numberOfPtsPerTraceType[type] += estimateDataLength(trace)
+    // other special cases
 
-    // cap the number of point per trace type
-    if (numberOfPtsPerTraceType[type] > maxPtsPerTraceType(type)) {
+    // box with boxpoints: 'all'
+    if (
+      type === 'box' &&
+      trace.boxpoints === 'all' &&
+      len > 5e4
```
alexcjohnson: In principle I guess the special cases should get included in the budget as well - otherwise you could include 199 x 5e4-point box traces...

src/component/plotly-graph/parse.js
```diff
@@ -155,11 +167,23 @@ function willFigureHang (result) {
...
+    if (innerArrays.length) {
+      return innerArrays
+        .map(a => a.length)
+        .reduce((a, v) => a + v)
+    } else {
+      return arr.length
+    }
+  })
+
+  return arrays.length ? Math.max(...lengths) : 0
```
alexcjohnson: 🐄 not that it matters, but `Math.max(0, ...lengths)` works too

src/component/plotly-graph/parse.js
etpinard: good call -> done in 
851b803

src/component/plotly-graph/parse.js
etpinard: Good idea -> 0372d82

src/component/plotly-graph/parse.js
etpinard: :cowboy_hat_face: -> 851b803

alexcjohnson:
Great, thanks! Still concerned about splom but we'll see once it exists 😄 
💃

etpinard:
> Still concerned about splom but we'll see once it exists

I have a good feeling splom with handle `1e7` points in less than 15 seconds. Aim high :rocket:

PR #59, opened 2018-03-06
Fetch plotlyJS bundle from source url

tarzzz:
Part of https://github.com/plotly/streambed/issues/10473

### DIFF
--- a/deployment/Dockerfile
+++ b/deployment/Dockerfile
@@ -115,4 +115,4 @@ COPY deployment/run_server /
 
 EXPOSE 9091
 ENTRYPOINT ["/run_server"]
-CMD ["--plotlyJS", "https://plot.ly/static/plotlyjs/build/plotlyjs-bundle.js --mathjax /mathjax/MathJax.js"]
+CMD ["--plotlyJS", "https://plot.ly/static/plotlyjs/build/plotlyjs-bundle.js", "--mathjax", "/mathjax/MathJax.js"]
--- a/deployment/Dockerfile
+++ b/deployment/Dockerfile
@@ -5,7 +5,7 @@ FROM ubuntu:xenial
 # From: https://github.com/nodejs/docker-node/blob/master/6.11/Dockerfile
 
 RUN apt-get update && apt-get install -y --no-install-recommends \
-        gnupg curl ca-certificates xz-utils wget \
+        gnupg curl ca-certificates xz-utils wget libgconf-2-4 \
     && rm -rf /var/lib/apt/lists/* && apt-get clean
 
 RUN groupadd --gid 1000 node \
@@ -115,4 +115,4 @@ COPY deployment/run_server /
 
 EXPOSE 9091
 ENTRYPOINT ["/run_server"]
-CMD ["--plotlyJS", "https://plot.ly/static/plotlyjs/build/plotlyjs-bundle.js", "--mathjax", "/mathjax/MathJax.js"]
+CMD ["--mathjax", "/mathjax/MathJax.js"]
--- a/deployment/run_server
+++ b/deployment/run_server
@@ -2,9 +2,16 @@
 
 fc-cache -v /usr/share/fonts/user
 
+BUILD_DIR=/var/www/image-exporter/build
+if [[ -n "${PLOTLY_JS_SRC}" ]]; then
+  # Fetch plotly js bundle and save it locally:
+  mkdir -p $BUILD_DIR
+  wget --retry-connrefused --no-check-certificate -O $BUILD_DIR/plotly-bundle.js $PLOTLY_JS_SRC
+  PLOTLYJS_ARG="--plotlyJS $BUILD_DIR/plotly-bundle.js"
+fi
+
 # Dynamically create the wrapper used to launch xvfb and the server so that
 # we can fill in $@ from the Docker "command" argument.
-
 cat > /usr/local/bin/xvfb_wrapper <<EOF
 #!/bin/bash
 
@@ -19,7 +26,7 @@ fi
 pkill Xvfb
 pkill node
 
-xvfb-run --auto-servernum --server-args '-screen 0 640x480x24' ./bin/plotly-export-server.js $@ 1>/proc/1/fd/1 2>/proc/1/fd/2 &
+xvfb-run --auto-servernum --server-args '-screen 0 640x480x24' ./bin/plotly-export-server.js $PLOTLYJS_ARG $@ 1>/proc/1/fd/1 2>/proc/1/fd/2 &
 echo \$! > \$PIDFILE
 
 EOF


tarzzz:
@etpinard Please review.. cc: @scjody

src/component/plotly-graph/inject.js
```diff
@@ -2,6 +2,9 @@ const fs = require('fs')
 const path = require('path')
 const isUrl = require('is-url')
 const semver = require('semver')
+const os = require('os')
+const curl = require('curlrequest')
```
etpinard: Did we really need another dep for this? Can the native `http` and `https` modules or even the more common [`request`](https://github.com/request/request) module do the job?

src/component/plotly-graph/inject.js
```diff
@@ -38,7 +41,7 @@ function inject (opts = {}) {
   }
 
   if (isNonEmptyString(plotlyJS)) {
-    let src = resolve(plotlyJS)
+    let src = resolvePlotlyJS(plotlyJS)
```
etpinard: I'd vote for adding an CLI / API option here. For example a boolean `--fetch-plotly.js` or more generally `--fetch-ressources` as some users might still want to leave it to the electron to app to fetch their custom (i.e. non-CDN) plotly.js bundles.

src/component/plotly-graph/inject.js
```diff
@@ -53,6 +56,28 @@ function inject (opts = {}) {
...
+      'output': path.join(os.tmpdir(), 'plotly-graph-exporter-build', 'plotly-bundle.js')
+    }
+
+    curl.request(curlOpts, function (err) {
+      if (err) {
+        throw new Error(`An error occured while retrieving PlotlyJS from: ${v}. Error: ${err}`)
+      }
+    })
+
+    return 'plotly-bundle.js'
```
etpinard: Hmm. So here you return a string before the request is finished. This could lead to a race condition where the browser window would open before the inject step is done fetching the plotly.js bundle. Sounds like we need to make the inject step async (just like parse, render and convert steps).

etpinard: It shouldn't be too hard to make the inject step async as [create-index.js](https://github.com/plotly/image-exporter/blob/1fae2a7efafe4d784a4dd996ec06956788b0ddaf/src/util/create-index.js) (i.e. where the component inject methods are called is already async.

scjody:
Another way to do this would be to run the `curl` command in the `run_server` script - that's what I had in mind when I suggested `curl`. But as long as the solution retries with exponential backoff and a maximum delay, I'm happy.

src/component/plotly-graph/inject.js
```diff
@@ -53,6 +56,28 @@ function inject (opts = {}) {
   return parts
 }
 
+function resolvePlotlyJS (v) {
+  if (isUrl(v)) {
+    // save the plotly-bundle to tmpdir and use it:
+    const curlOpts = {
+      'url': v,
+      'retries': 10,
```
scjody: Can you explain what happens if plotly.js can't be retrieved within 10 retries?

etpinard:
> Another way to do this would be to run the curl command in the run_server script - that's what I had in mind when I suggested curl

That sounds to me like a more robust solution :+1:

tarzzz:
Looking at the initial comments, it seems using curl directly in`run_server` and adding a `--fetch` CLI argument makes more sense.. I can try that approach .. !! 

> Can you explain what happens if plotly.js can't be retrieved within 10 retries?

As per the current code, it will throw an Error. The server would still start, but it wont be working. Therefore I think adding it in `run_server` makes more sense, because then container will explicitly Fail.

scjody:
The image-exporter can already use a local `plotly.js` (retrieved by `curl` in this case), can't it? I don't understand the need for `--fetch`, unless I'm mistaken here.

tarzzz:
`--fetch` is to address @etpinard 's [comment](https://github.com/plotly/image-exporter/pull/59#discussion_r172537412):

> I'd vote for adding an CLI / API option here. For example a boolean --fetch-plotly.js or more generally --fetch-ressources as some users might still want to leave it to the electron to app to fetch their custom (i.e. non-CDN) plotly.js bundles.

scjody:
That comment assumes that the server is doing the fetch though, doesn't it? If the server just has to use a local `plotly.js`, what would the `--fetch` option do?

etpinard:
No need for a CLI / API option if we're going with the -curl directly in run_server- route.

deployment/run_server
```diff
@@ -1,10 +1,34 @@
...
+
+while [[ $# -gt 0 ]]
+  do
+    key="$1"
+    value="$2"
+    # Check for valid url and update args:
+    if [[ $key = "--plotlyJS" ]] && [[ $value =~ $URL_REGEXP ]]; then
+
+      mkdir -p $BUILD_DIR
+      wget --tries 40 --retry-connrefused -O $BUILD_DIR/plotly-bundle.js $value
```
tarzzz: using wget instead of `curl` because it has `--retry-connrefused` option. Curl also has this option in newer versions, which requires an upgrade. This works as it is.

I noticed that it takes ~10 retries to fetch the bundle on buildly , so keeping it to 40 to be on the safer side.

deployment/Dockerfile
```diff
@@ -115,4 +115,4 @@ COPY deployment/run_server /
 
 EXPOSE 9091
 ENTRYPOINT ["/run_server"]
-CMD ["--plotlyJS", "https://plot.ly/static/plotlyjs/build/plotlyjs-bundle.js --mathjax /mathjax/MathJax.js"]
+CMD ["--plotlyJS", "https://plot.ly/static/plotlyjs/build/plotlyjs-bundle.js", "--mathjax", "/mathjax/MathJax.js"]
```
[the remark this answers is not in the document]
etpinard: Brilliant :sparkles:

deployment/run_server
```diff
@@ -1,10 +1,34 @@
 #!/bin/bash -xe
 
 fc-cache -v /usr/share/fonts/user
+set -e
+
+BUILD_DIR=/tmp/plotly-graph-exporter-build
```
etpinard: I'm curious, why start from `/tmp/` ? This repo has already a `build/` directory, perhaps that would be cleaner?

etpinard:
@tarzzz your latest patches look good to me, but `deployment/` is @scjody's work so he's the main reviewer here.

deployment/run_server
scjody: Can you explain the consequences if this download fails?

deployment/run_server
```diff
@@ -1,10 +1,34 @@
...
+URL_REGEXP='(https?|ftp|file)://[-A-Za-z0-9\+&@#/%?=~_|!:,.;]*[-A-Za-z0-9\+&@#/%=~_|]'
+ARGS=''
+
+while [[ $# -gt 0 ]]
```
scjody: Do you need something this complicated? This looks fragile, for example if we need to pass single arguments like `--debug`.

Alternatives:
* Pass the plotly.js URL as an environment variable. If the variable is set, `run_server` does the download and adds the appropriate `--plotlyJS PATH` arguments to another variable (e.g. `PLOTLYJS_ARG`). Then you can run the server as `[...] ./bin/plotly-export-server.js $@ $PLOTLYJS_ARG [...]`.
* Pass the plotly.js URL as the first argument, or `--` if there isn't one. If the argument is not `--`, run_server does the download and adds the path arguments as above.

tarzzz: We can fix this by using only one ‘shift’ in the ‘else’ block, but i am also okay with reserving ‘$1’ for plotlyJS argument.. (which can be a url, version-tag or ‘—‘)

scjody: The one `shift` approach essentially duplicates most of `plotly-export-server.js`'s argument parsing code into `run_server`, which is why I'm not in favour of it.

Is there any reason not to go with the environment variable-based approach? That seems likely to be the most reliable approach since it doesn't involve parsing the argument string in any way.

tarzzz: Environment Var based approach sounds okay to me.. !! We can also remove 'plotlyJS' arg from Dockerfile.. !!

deployment/run_server
tarzzz: It fails silently. imageserver will not start, but the other onprem services go on as usual.

This is because we report streambed's `/ping` as ready-state. Maybe we can add imageserver's `ping` as well? i.e. if we want the service to error out on imageserver start failure.

scjody:
> It fails silently. imageserver will not start, but the other onprem services go on as usual.

Please come up with a solution that lets us keep trying until startup succeeds. Failing silently is not a good option (in this case, and in most cases).

If the imageserver container exits, it will be restarted until it succeeds, so that could help. Or you could use monit, which is already in the imageserver container.

tarzzz:
> Please come up with a solution that lets us keep trying until startup succeeds. Failing silently is not a good option (in this case, and in most cases).

I checked by adding an incorrect url in replicated (ie. which gives 404). So, the imageserver container will restart on failure. I was wondering whether the replicated dashboard should show the status for On Premise sevice as 'Started' or 'Starting' when this happens.. !!

tarzzz:
Unless there is something wrong which prevents image-server from downloading the `plotly-bundle.js` **at all**,  this current approach will ensure that image-server is started properly (even if it takes multiple retries / container restarts).. !!

tarzzz:
@scjody Please review the updates.. !! ( I am not sure if you have already 👀 them, but wanted to confirm)

deployment/run_server
```diff
@@ -1,10 +1,25 @@
 #!/bin/bash -xe
 
 fc-cache -v /usr/share/fonts/user
+set -e
```
scjody: This is already set at the top of the file.

deployment/run_server
```diff
@@ -1,10 +1,25 @@
...
+
+BUILD_DIR=/var/www/image-exporter/build
+URL_REGEXP='(https?|ftp|file)://[-A-Za-z0-9\+&@#/%?=~_|!:,.;]*[-A-Za-z0-9\+&@#/%=~_|]'
```
scjody: This is unneeded now. Just pass an environment variable that, if set, contains a URL. That will be significantly more maintanable than depending on a regular expression.

scjody: In other words there should be two modes:
* Retrieve plotly.js from a URL and pass that. In this case we download and add the argument.
* The caller specifies the `--plotlyJS` argument.

deployment/run_server
```diff
@@ -1,10 +1,25 @@
...
+
+if [ ! -z ${PLOTLY_JS_SRC+x} ]; then
+  if [[ $PLOTLY_JS_SRC =~ $URL_REGEXP ]]; then
+    # Fetch plotly js bundle and save it locally:
+    mkdir -p $BUILD_DIR
+    wget --retry-connrefused --no-check-certificate -O $BUILD_DIR/plotly-bundle.js $PLOTLY_JS_SRC
+    PLOTLYJS_ARG="--PlotlyJS $BUILD_DIR/plotly-bundle.js"
+  else
+    # Pass on the exact arguments:
+    PLOTLYJS_ARG="--PlotlyJS $PLOTLY_JS_SRC"
```
scjody: Note that the argument is `--plotlyJS`. Perhaps this works as well but please use the correct argument for consistency.

tarzzz:
@scjody Updates made.. !!

deployment/run_server
```diff
@@ -2,9 +2,16 @@
 
 fc-cache -v /usr/share/fonts/user
 
+BUILD_DIR=/var/www/image-exporter/build
+if [ ! -z ${PLOTLY_JS_SRC+x} ]; then
```
scjody: I'm unclear on the purpose of `+x`. Can you please explain?

tarzzz: https://stackoverflow.com/a/13864829

tarzzz: There shouldn’t be a case when the variable is set to blank string, but this is more robust.. !!

scjody: OK, but in this case if `PLOTLY_JS_SRC` is set tho the empty string we also want to skip the download, so the simpler approach is preferable:

```bash
if [[ -n "$PLOTLY_JS_SRC" ]]; then
    ....
fi
```

Note that I'm also using `[[` for reasons outlined in https://github.com/progrium/bashstyle

tarzzz: > but in this case if PLOTLY_JS_SRC is set tho the empty string we also want to skip the download,

That makes sense to me.. updated .. !!

PR #60, opened 2018-03-15
Add Dash Apps Preview + PDF Generation

tarzzz:
**WIP**


src/component/plotly-dash-preview/parse.js
```diff
@@ -0,0 +1,39 @@
...
+  } else {
+    return errorOut(400, 'invalid url')
+  }
+
+  if (isNonEmptyString(body.format) && (body.format === 'png' || body.format === 'pdf')) {
+    result.format = body.format
+  } else {
+    // Default
+    result.format = 'png'
+  }
```
tarzzz: We don't need to support other image types, since these are only going to be used for PDF exports and Organize thumbnails ..

src/component/plotly-dash-preview/render.js
```diff
@@ -0,0 +1,57 @@
...
+      result.msg = 'dash preview generation failed'
+    }
+    sendToMain(errorCode, result)
+  }
+
+  win.on('closed', () => {
+    win = null
+  })
+
+  contents.on('page-favicon-updated', () => {
```
tarzzz: This is a placeholder for now .. !! It will be replaced with a custom event fired by Dash App once the rendering has finished.

src/component/plotly-dash-preview/render.js
```diff
@@ -0,0 +1,57 @@
...
+    if (PRINT_TO_PDF) {
+      contents.printToPDF({}, (err, pdfData) => {
```
tarzzz: TODO Error handling.. !!

tarzzz: Also, the PDF Options is set to `{}` until we have some more info from: https://github.com/plotly/streambed/issues/10698#issuecomment-373161617 ..

tarzzz:
@etpinard Can I please get an initial review on this.. I'll be adding some tests in the meantime.. !!

test/unit/plotly-dash-preview_test.js
```diff
@@ -0,0 +1,120 @@
...
+    sinon.stub(remote, 'createBrowserWindow').returns(win)
+    win.webContents.printToPDF.yields(null, '-> image data <-')
+
+    fn({
+      width: 500,
+      height: 500,
+      url: 'dummy',
+      format: 'pdf'
+    }, {}, (errorCode, result) => {
+      t.ok(win.webContents.printToPDF.calledOnce)
```
tarzzz: This fails because `printToPDF` waits for `page-favicon-updated` event before being called. Is there a way to handle this? The next tests also fail for similar reason.. !!

src/component/plotly-dash-preview/render.js
chriddyp: Could we add these options to the API? https://github.com/electron/electron/blob/master/docs/api/web-contents.md#contentsprinttopdfoptions-callback

src/component/plotly-dash-preview/render.js
```diff
@@ -0,0 +1,62 @@
...
+            done(525)
+        }
+        else {
+          result.imgData = pdfData
+          done()
+        }
+      })
+    } else {
+      contents.capturePage(img => {
+        result.imgData = img.toPNG()
```
chriddyp: interesting, i'm curious what a PNG looks like! this could be very cool for thumbnails 😸

src/component/plotly-dash-preview/parse.js
```diff
@@ -0,0 +1,39 @@
...
+
+  result.width = body.width || 800
+  result.height = body.height || 600
```
chriddyp: This is interesting, I didn't think about this. I'm assuming that with PDF reports, where the content has absolute pixel sizing (e.g. the pages are 8.5" wide and 11" long), it doesn't matter what width and height we provide. However, for responsive apps, we'll need to provide something reasonable. Let's try a medium sized laptop: `width: 1680px`, `height: 1050`

src/component/plotly-dash-preview/render.js
tarzzz: Yes, that's the plan. I did test it with the vanguard app:

a

It only captures the visible area in the browser, hence the need for `width`/`height` options ..

src/component/plotly-dash-preview/parse.js
tarzzz: These options are only meant for `png` generation, as the printToPDF outputs are determined by: https://github.com/electron/electron/blob/master/docs/api/web-contents.md#contentsprinttopdfoptions-callback .. and not be the browser-size.

Although, we _can_ add them in printToPDF options after some conversion as `printToPDF` expects height/width in microns.

chriddyp: Ah I understand now. OK, lets just use the electron options directly instead of doing a pixel to micron conversion.

etpinard:
Looking good so far. Glad someone else is trying to add things to image-expoter. That was the goal all along.

Now, by preview you mean combining reports and thumbnail generation in one component? Electron's capturePage and printToPDF work very differently and support different options. So, I'd vote for splitting this thing into two separate components.

test/unit/plotly-dash-preview_test.js
etpinard: `win` is an event emitter as created [here](https://github.com/plotly/image-exporter/blob/934999da692e2cdc61fdf7debdb87e022330f640/test/common.js#L25-L43), so calling `win.emit('page-favicon-updated')` should allow you to test the callback.

tarzzz:
> Now, by preview you mean combining reports and thumbnail generation in one component?

Yes, that was the my initial plan.. 

>  So, I'd vote for splitting this thing into two separate components.

In that case, I will remove the `capturePage` for now (if it seems ok to you), since we only need PDF generation for now (right @chriddyp ? ). I can add thumbnails after `2.4.0` release.

chriddyp:
>  > So, I'd vote for splitting this thing into two separate components.

> In that case, I will remove the capturePage for now (if it seems ok to you), since we only need PDF generation for now

yeah, we only need PDF generation for now

tarzzz:
@etpinard @chriddyp: Some comments/updates:
  - we are using 'waitfor' div in dash-apps to determine if the app has finished rendering (before taking snapshot)
  - printToPDF's pdfOptions will be present in the request body and supplied to `printToPDF`.

@etpinard Please review.. 
The node (V8) tests seem to be failing for reasons unrelated to this PR.. would be great if you could take a look at them.. !!

src/component/plotly-dash-preview/convert.js
```diff
@@ -0,0 +1,22 @@
...
+ * @param {object} opts : component options
+ * @param {function} reply
+ *  - errorCode
+ *  - result
+ */
+function convert (info, opts, reply) {
+  const result = {}
+
+  result.head = {}
+  result.head['Content-Type'] = cst.contentFormat['pdf']
```
etpinard: :cow2: `cst.contentFormat.pdf` would suffice.

src/component/plotly-dash-preview/index.js
```diff
@@ -0,0 +1,10 @@
+const plotlyGraph = require('../plotly-graph')
+
+module.exports = {
+  name: 'plotly-dash-preview',
+  ping: require('../../util/generic-ping'),
+  inject: plotlyGraph.inject,
```
etpinard: No need for an `inject` method here as the `render` step has no external deps.

src/component/plotly-dash-preview/parse.js
```diff
@@ -0,0 +1,30 @@
...
+ */
+function parse (body, opts, sendToRenderer) {
+  const result = {}
+
+  const errorOut = (code, msg) => {
+    result.msg = msg
+    sendToRenderer(code, result)
+  }
+
+  if (isNonEmptyString(body.url)) {
```
etpinard: We could probably be more strict here. Using `isUrl` (like done [here](https://github.com/plotly/image-exporter/blob/a7e67453807aa08c3b27dbdfb0e81934eab5c134/src/component/plotly-graph/inject.js#L57)) would be best.

src/component/plotly-dash-preview/parse.js
```diff
@@ -0,0 +1,30 @@
...
+    result.url = body.url
+  } else {
+    return errorOut(400, 'invalid url')
+  }
+
+  result.pdfOptions = body.pdf_options || {}
```
etpinard: What does Electron do when we pass invalid pdf options to `printToPDF`? 

- Does it ignore bad values and use the defaults?
- Does it error out?

if the latter, we should try to intercept bad pdf options here, so that we return a 400-series parse error instead of 500-series render error.

etpinard: ... if the former, then we don't need to do anything here :tada:

bin/plotly-export-server_electron.js
```diff
@@ -53,6 +53,11 @@ const opts = {
       name: 'plotly-dashboard-preview',
       route: '/dashboard-preview',
       options: plotlyJsOpts
+    },
+    {
+      name: 'plotly-dash-preview',
+      route: '/dash-preview',
+      options: plotlyJsOpts
```
etpinard: No need to pass plotly.js options here. You can :hocho: this line.

src/component/plotly-dash-preview/render.js
```diff
@@ -0,0 +1,65 @@
...
+ * @param {object} opts : component options
+ * @param {function} sendToMain
+ *  - errorCode
+ *  - result
+ *    - imgData
+ */
+function render (info, opts, sendToMain) {
+  const result = {}
+
+  let win = remote.createBrowserWindow()
```
etpinard: No need to specify the window width/height? That is, you're assuming all dash reports will have the same dimensions?

src/component/plotly-dash-preview/render.js
```diff
@@ -0,0 +1,65 @@
...
+        contents.printToPDF(info.pdfOptions, (err, pdfData) => {
+          if (err) {
+            done(525)
+          } else {
+            result.imgData = pdfData
+            done()
+          }
+        })
+      }
+    })
```
etpinard: What if `finishedLoaded` doesn't resolve?

src/component/plotly-dash-preview/render.js
```diff
@@ -0,0 +1,65 @@
...
+          resolve(true);
+        }
+      })`)
+  }
+
+  win.on('closed', () => {
+    win = null
+  })
+
+  let intervalId = setInterval(() => {
```
etpinard: I'm curious. Why did you decide to place the `setInterval` in the main window scope instead of in the child window scope?

Would something like:

```js
const loaded = () => {
  return contents.executeJavaScript(`
     new Promise((resolve, reject) => {
       let tries = 10
       let interval = setInterval(() => {
         const el = document.getElementById('waitfor')
         if (el) {
           clearInterval(interval)
           resolve(true)
         }
         if (--tries === 0) reject('fail to load')
       }, 500)
     })
  `) 
}

loaded().then(() => {
   contents.printToPDF(info.pdfOptios, (err, pdfData) => {
      if (err) {
        done(525)
      } else {
        result.imgData = pdfData
        done()
      }
   }).catch(() => {
     done(525)
   })
})
```

test/unit/plotly-dash-preview_test.js
```diff
@@ -0,0 +1,72 @@
...
+      t.equal(errorCode, undefined, 'code')
+      t.equal(result.imgData, '-> image data <-', 'result')
+      t.end()
+    })
+  })
+
+  t.test('should handle printToPDF errors', t => {
+    const win = createMockWindow()
+    sinon.stub(remote, 'createBrowserWindow').returns(win)
+    win.webContents.executeJavaScript.resolves(true)
```
etpinard: Nice tests! Like mentioned in an earlier comment, we'll need to handle (and test :smirk: ) the case where `executeJavascript` does not resolve.

src/component/plotly-dash-preview/render.js
chriddyp: Some discussion on this earlier here: https://github.com/plotly/image-exporter/pull/60#discussion_r174827073

Now that I think about it, I wonder if we should still fix the size of the window before specifying the size of the PDF export. The PDF Export size will likely change the viewport (so responsive CSS will be reflected in the PDF) but I doubt that it will re-renderer all of the javascript. So, if there are certain components that are responsive with JS and not CSS, then I doubt that they will be reflected at the new size.

So, I think it makes sense to set size of this window to be the same size as requested in the PDF renderer. For printed content, 1px will be 1/96th of an inch (https://www.w3.org/Style/Examples/007/units.en.html). So 8.5 x 11 is 816px x 1056

src/component/plotly-dash-preview/render.js
```diff
@@ -0,0 +1,65 @@
...
+  }
+
+  /*
+   * We check for a 'waitfor' div in the dash-app
+   * which indicates that the app has finished rendering.
+   */
+  const finishedLoading = () => {
+    return win.webContents.executeJavaScript(`
+      new Promise((resolve, reject) => {
+        var a = document.getElementById("waitfor");
```
chriddyp: Can we make this a variable? i.e. part of the API where the use can pass in an arbitrary CSS selector. That would modify this to be `document.querySelector`. 

For example, if the app has a graph in it, I might end up using a query selector like `#mygraph .plot-container` in order to wait until the graph was rendered (as plotly.js draws `.plot-container` dynamically). But, if the report doesn't have a graph in it, then I'd inject my own div with id `waitfor`

tarzzz: Just to confirm, the query selector comes as the part of the incoming POST request (with `.plot-container` as default if nothing is provided)? 

It should be safe to assume the dash-app has atleast one chart even if the incoming request does not specify it.

src/component/plotly-dash-preview/render.js
tarzzz: Elaborating on ⬆️ my point was that:

> A4 paper measures in at 210mm by 297mm (8.26" by 11.69") whereas US Letter paper measures at 8.5" by 11" (215.9mm by 279.4mm)

Since difference is not too much between A4 and Letter, it shouldn't really affect the viewports if we change it by the `paperSize` (though we do get a significant difference when we consider A3/A2 as well)

If we consider A2, it comes out to ~`1584.0px x 2246px` .. which is big compared to general screen-sizes. (or `xvfb` screen sizes in case of headless server)

src/component/plotly-dash-preview/parse.js
tarzzz: > Does it ignore bad values and use the defaults?

That's correct. No 500s ..

tarzzz:
@etpinard @chriddyp Updates made:

  - Atleast one of `loading_selector` or `timeout` should be specified in request. If `timeout` is specified, we wait for this time before capturing the page. 
  - `pageSize` should be specified either as the part of the request `body` or as the part of `pdfOptions`. We use it to determine the browser window size as indicated by Chris in https://github.com/plotly/image-exporter/pull/60#discussion_r177200389
  - Added tests for these ⬆️ options as well as to check for rejected promises (`executeJavascript`.
  - Addressed minor code-style related comments.

src/component/plotly-dash-preview/parse.js
```diff
@@ -24,6 +24,13 @@ function parse (body, opts, sendToRenderer) {
   }
 
   result.pdfOptions = body.pdf_options || {}
+  if (!body.loading_selector && !body.timeout) {
```
etpinard: We should use utils `isPositiveNumber` for `timeout` and `isNonEmptyString` for `loading_selector`.

Oh and I would vote for just `selector` instead of `loading_selector` which is unnecessarily verbose here I think.

src/component/plotly-dash-preview/parse.js
etpinard: Ok that's great. We won't have to worry about the various `printToPDF` options :ok_hand:

src/component/plotly-dash-preview/render.js
```diff
@@ -31,35 +33,53 @@ function render (info, opts, sendToMain) {
...
+  const loaded = () => {
     return win.webContents.executeJavaScript(`
       new Promise((resolve, reject) => {
-        var a = document.getElementById("waitfor");
-        if (a) {
-          resolve(true);
+        let tries = ${cst.maxRenderingTries}
+
+        if (${info.timeOut}) {
+          tries = parseInt(${info.timeOut} * 1000/${cst.minInterval})
```
etpinard: Let's move the coercion step to `parse`. Note that `Number(info.timeOut)` would suffice. `parseInt` is slow.

etpinard: :cow2: `1000 / ${cst.minInterval)` please.

src/component/plotly-dash-preview/render.js
```diff
@@ -31,35 +33,53 @@ function render (info, opts, sendToMain) {
...
         }
+
+        let interval = setInterval(() => {
+          let el = document.querySelector("${info.loadingSelector}")
```
etpinard: :cow2: use `'` over `"` even inside \`\`

src/component/plotly-dash-preview/render.js
```diff
@@ -31,35 +33,53 @@ function render (info, opts, sendToMain) {
...
+      if (err) {
+        done(525)
+      } else {
+        result.imgData = pdfData
+        done()
       }
     })
-  }, 500)
+  }).catch(() => {
+    done(525)
```
etpinard: Sorry to not having noticed this before. But, it would be nice to separate `printToPDF` errors (that's the `done(525)` inside the `printToPDF` callback) and this one which is the result of a timeout. 

I propose making `525` the `printToPDF` callback error, and adding `526` when we hit a timeout.

package.json
```diff
@@ -49,7 +49,7 @@
     "get-stdin": "^5.0.1",
     "glob": "^7.1.2",
     "is-plain-obj": "^1.1.0",
-    "is-url": "^1.2.2",
+    "is-url": "1.2.2",
```
etpinard: Ok yeah. That might work. But really, we should be using the latest version of `is-url`. Looking at their [changelog](https://github.com/segmentio/is-url/commits/master), bumping to version 1.2.4 should fix everything.

src/component/plotly-dash-preview/parse.js
```diff
@@ -31,6 +32,19 @@ function parse (body, opts, sendToRenderer) {
...
+    result.browserSize = cst.sizeMapping[result.pdfOptions.pageSize]
+  } else if (body.pageSize && body.pageSize.width && body.pageSize.height) {
+    result.browserSize = {
+      width: body.pageSize.width * cst.pixelsInMicron,
+      height: body.pageSize.height * cst.pixelsInMicron
+    }
+  } else {
+    return errorOut(400, 'pageSize must either be A3, A4, A5, Legal, Letter, ' +
+                         'Tabloid or an Object containing height and width ' +
+                         'in microns.')
```
etpinard: :cow2: I really see indents like this in JS. That seems very pythonic. I would write this as:

```js
errorOut(
  400,
  'pageSize must either be A3, A4, A5, Legal, Letter, ' +
  'Tabloid or an Object containing height and width ' +
  'in microns.'
)
```

src/component/plotly-dash-preview/parse.js
```diff
@@ -31,6 +32,19 @@ function parse (body, opts, sendToRenderer) {
   result.loadingSelector = body.loading_selector
   result.timeOut = body.timeout
 
+  if (cst.sizeMapping[result.pdfOptions.pageSize]) {
+    result.browserSize = cst.sizeMapping[result.pdfOptions.pageSize]
+  } else if (body.pageSize && body.pageSize.width && body.pageSize.height) {
```
etpinard: Use `isPositiveNumeric(body.pageSize.width)` and `isPositiveNumeric(body.pageSize.height)` to error early on 0 and negative inputs.

test/unit/plotly-dash-preview_test.js
```diff
@@ -23,6 +24,28 @@ tap.test('parse:', t => {
...
+    
+    fn({url: 'https://dash-app.com'}, {}, (errorCode, result) => {
+      t.equal(errorCode, 400)
+      t.equal(result.msg, 'either loading_selector or timeout must be specified')
+      t.end()
+    })
+  })
+
+  t.test('should error when pageSize is not given', t => {
+
```
etpinard: :zap:

test/unit/plotly-dash-preview_test.js
```diff
@@ -23,6 +24,28 @@ tap.test('parse:', t => {
     t.end()
   })
 
+  t.test('should error when neither loading_selector or timeout is given', t => {
+
```
etpinard: :zap:

src/component/plotly-dash-preview/parse.js
scjody: It would be less frustrating to the user if we could return a 400-series error (and have that bubble up through streambed). For example if a user is trying to set the page size they may try `'page_size': 'legal'`, get back a letter option (because that's the default), and keep trying variations of `legal`. If we instead sent a 400 error saying that `page_size` is an invalid option, that will be clearer to the user.

If this is a lot of work, I don't feel strongly about it though.

src/component/plotly-dash-preview/render.js
tarzzz:
``` 
526: 'plotly.js version 1.11.0 or up required',
```
526 is already taken.. maybe traditional `408 Request Timeout` or `504 Gateway Timeout`?

src/component/plotly-dash-preview/parse.js
etpinard: I agree. Returning a 400-series error here would be best from the user's standpoint. But since we would have to duplicate how Electron validates input here in the parse step in order to get this right, this may turn into a developer's nightmare for little gain. I'd vote for leaving this as is.

src/component/plotly-dash-preview/parse.js
```diff
@@ -25,24 +27,29 @@ function parse (body, opts, sendToRenderer) {
   }
 
   result.pdfOptions = body.pdf_options || {}
-  if (!body.loading_selector && !body.timeout) {
-    return errorOut(400, 'either loading_selector or timeout must be specified')
+  if (!isNonEmptyString(body.selector) && !isPositiveNumeric(body.timeout)) {
+    return errorOut(400, 'either selector or timeout must be specified')
```
etpinard: This here may be overly stringent. Perhaps we could default to a 3000ms timeout if `timeout` and `selector` are omitted from the POST request. @chriddyp what do you think?

src/component/plotly-dash-preview/render.js
```diff
@@ -35,14 +35,10 @@ function render (info, opts, sendToMain) {
   const loaded = () => {
     return win.webContents.executeJavaScript(`
       new Promise((resolve, reject) => {
-        let tries = ${cst.maxRenderingTries}
-
-        if (${info.timeOut}) {
-          tries = parseInt(${info.timeOut} * 1000/${cst.minInterval})
-        }
+        let tries = info.tries || ${cst.maxRenderingTries}
```
etpinard: Very clean here. Nice :ok_hand:

src/component/plotly-dash-preview/render.js
etpinard: It's already taken in the `plotly-graph` component, right? Not in `plotly-dash-preview`? I don't like overriding 504 (a gateway timeout is something else) and I want this to be a 500-series.

src/component/plotly-dash-preview/parse.js
scjody: OK, if it's easy let's do it but I agree that this isn't worth it if it's painful or a lot of work.

tarzzz:
Added integration test, and ran coverage:
image

tarzzz:
Updates: Added integration test, and `timeout` code (526) for rendering related timeouts.

Still blocked by intermittently failing integration tests..

src/component/plotly-dash-preview/render.js
```diff
@@ -23,7 +23,7 @@ function render (info, opts, sendToMain) {
     win.close()
 
     if (errorCode) {
-      result.msg = 'dash preview generation failed'
+      result.msg = cst.statusMsg[errorCode]
```
etpinard: nicely done :+1:

etpinard:
@tarzzz thanks for running `npm run coverage`. You can ignore the uncovered lines in `plotly-dash-preview/convert.js`. But it would be nice to bring up `plotly-dash-preview/parse.js` to 100% coverage.

As for our plotly-export-server integration tests. Perhaps we should skip them on CI for now (and open a new issue about it). I think booting up 6 browser windows on the CI machines might be a little too much for them to handle. We could revisit this later.

tarzzz:
@etpinard : Coverage ..

image

Added issue to address intermittent errors: https://github.com/plotly/image-exporter/issues/61

PR #62, opened 2018-03-29
Dash App Preview generation

tarzzz:


src/component/plotly-dash-preview/parse.js
```diff
@@ -0,0 +1,58 @@
...
+    return errorOut(400, 'either selector or timeout must be specified')
+  }
+
+  result.selector = body.selector
+  result.timeOut = body.timeout
+  result.tries = Number(result.timeOut * 1000 / cst.minInterval)
+
+  if (cst.sizeMapping[result.pdfOptions.pageSize]) {
+    result.browserSize = cst.sizeMapping[result.pdfOptions.pageSize]
+  } else if (body.pageSize && isPositiveNumeric(body.pageSize.width) &&
```
etpinard: All right. Looks like https://github.com/plotly/image-exporter/pull/60#discussion_r177752390 didn't get addressed. Just double-check with @tarzzz if this is on purpose.

Personally, making users have to specify a `selector` string or a `timeout` interval AND a `pageSize` just to get an export feels wrong. 

Wouldn't it be great if `{url: "server url", method: 'POST', body: '{"url": "dash-app-url"}'}` would _just work_<sup>TM</sup> (most of the time).

tarzzz:
Thanks @etpinard  ..

PR #63, opened 2018-03-30
Enable CI platforms to build Linux, Windows and OS X releases

JamesCropcho:
Hello all,

This would enable the CI platforms to build a cross-platform set of releases:
* https://ci.appveyor.com/project/JamesCropcho/image-exporter-vm18n/history
* https://circleci.com/gh/plotly/image-exporter
* https://travis-ci.org/plotly/image-exporter/builds

These releases have not been tested on their respective platforms, nor are they necessarily in their final format. However, this pull would indeed enable _some form of_ successful build for each of the three platforms we are targeting.

What do you all think?

―James


.circleci/config.yml
```diff
@@ -29,7 +33,7 @@ jobs:
       - run:
           name: List versions
           command: |
-            echo "npm: $(npm --version)"
+            echo "yarn: $(yarn --version)"
```
etpinard: I'd prefer sticking to `npm` unless yarn has clear upsides.

.circleci/config.yml
```diff
@@ -39,40 +43,39 @@ jobs:
...
     <<: *base
     docker:
-      - image: circleci/node:6.10.3-browsers
+      - image: circleci/node:6-browsers
 
   test-node-v8:
     <<: *base
     docker:
-      - image: circleci/node:8.4.0-browsers
+      - image: circleci/node:8-browsers
```
etpinard: Why?

package.json
```diff
@@ -13,27 +13,43 @@
     "test:lint": "standard | snazzy",
     "test:unit": "tap test/unit/*_test.js",
     "test:integration": "tap test/integration/*_test.js",
-    "test": "npm run test:lint && npm run test:unit && npm run test:integration",
-    "coverage": "npm run test:unit -- --cov",
-    "lint": "standard --fix"
+    "test": "yarn run test:lint && yarn run test:unit && yarn run test:integration",
+    "coverage": "yarn run test:unit -- --cov",
+    "lint": "standard --fix",
+    "pack": "cross-env NODE_ENV=production electron-builder --publish=never"
```
etpinard: So this works? Nice!

.circleci/config.yml
etpinard: ... though you adding a `packing-lock.json` file to this project should help us trigger consistent builts on CI.

package.json
```diff
@@ -44,6 +60,7 @@
   ],
   "dependencies": {
     "body": "^5.1.0",
+    "cross-env": "^5.1.3",
```
n-riesco: If I'm not mistaken, cross-env should go devDependencies.

package.json
```diff
@@ -13,27 +13,43 @@
...
+    ],
     "appImage": {
-      "systemIntegration": "doNotAsk",
-      "category": "Utility"
+      "systemIntegration": "doNotAsk"
+    },
+    "linux": {
+      "category": "Utility",
+      "executableName": "plotly-graph-exporter",
+      "maintainer": "enquiries@nicolasriesco.net",
```
n-riesco: @etpinard I guess that here should go your email.

package.json
```diff
@@ -13,27 +13,43 @@
...
+      "target": [
```
n-riesco: @JamesCropcho I guess you want to add `"AppImage",` here.

.circleci/config.yml
```diff
@@ -106,3 +108,17 @@ jobs:
...
+    steps:
+      - checkout
+      - run:
+          name: Install deps
+          command: yarn install
+      - run:
+          name: Electron-Builder pack
+          command: yarn run pack
+      - store_artifacts:
+          path: release
```
n-riesco: We don't really want to include `release/linux-unpacked` (131MB).
See how Falcon does it:
https://github.com/plotly/falcon-sql-client/blob/3ebfd4d20a14973f6023990fabfff0cc285e0b21/circle.yml#L37-L41

.circleci/config.yml
```diff
@@ -19,6 +19,10 @@ workflows:
             - test-node-v6
             - test-node-v8
             - docker-build
+      - electron-pack-and-release:
```
n-riesco: This job didn't run:

image

.circleci/config.yml
```diff
@@ -29,7 +33,7 @@ jobs:
       - run:
           name: List versions
           command: |
-            echo "npm: $(npm --version)"
+            echo "yarn: $(yarn --version)"
             echo "node: $(node --version)"
       - checkout
```
n-riesco: In Falcon, we also keep a cache for `.git`.
See https://github.com/plotly/falcon-sql-client/blob/3ebfd4d20a14973f6023990fabfff0cc285e0b21/circle.yml#L12-L21

.circleci/config.yml
```diff
@@ -39,40 +43,39 @@ jobs:
...
       - run:
           name: List deps
           command: |
-            npm ls
+            yarn list
             pdftops -v
       - save_cache:
           paths:
             - node_modules
           key: v{{ .Environment.CIRCLE_CACHE_VERSION }}-deps-{{ .Branch }}-{{ .Environment.CIRCLE_JOB }}-{{ checksum "package.json" }}
```
n-riesco: If I'm not mistaken using `{{ .Environment.CIRCLE_JOB }}` defeats the purpose of having a cache.
Falcon  uses `key: v2-yarn-{{ checksum "yarn.lock" }}`.
`key: v2-yarn-{{ checksum "package-lock.json" }}` should work with `npm`.

.circleci/config.yml
n-riesco: @JamesCropcho BTW, `Dockerfile` and `deployment/Dockerfile` are still using `npm`.

.circleci/config.yml
```diff
@@ -106,3 +108,17 @@ jobs:
             docker tag quay.io/plotly/image-exporter:$CIRCLE_SHA1 quay.io/plotly/image-exporter:$CIRCLE_BRANCH
             docker push quay.io/plotly/image-exporter:$CIRCLE_SHA1
             docker push quay.io/plotly/image-exporter:$CIRCLE_BRANCH
+
+  electron-pack-and-release:
+    docker:
+      - image: circleci/node:8-browsers
+    steps:
```
n-riesco: I'd add the same steps to save and restore caches for `.git` and `node_modules` as those in the job `test-node-v8`.

appveyor.yml
```diff
@@ -0,0 +1,34 @@
...
+
+platform:
+  - x64
+
+#services:
+
+cache:
+  - '%USERPROFILE%\.electron'
+  - C:\Users\appveyor\AppData\Local\electron-builder\cache
+  - C:\projects\image-exporter-vm18n\node_modules
```
n-riesco: @JamesCropcho I don't know why these paths are listed in the cache. Have you found any documentation about this?

I've noticed `electron-builder` uses [these paths](https://github.com/electron-userland/electron-builder/blob/8f19ea99a9cc6fa4796f8057c268541a7b7f63b4/appveyor.yml#L4-L7)

n-riesco:
@JamesCropcho @etpinard I've tested the installers and they don't work. To get them working:
- we need to disable `asar`
- [getExportedArgs](https://github.com/plotly/image-exporter/blob/1c164bb2a4cb69f67a0f842aaf371bcb7721ca55/bin/args.js#L99) should use `slice(1)` instead of `slice(2)` when the executable from the installer is used.

.circleci/config.yml
JamesCropcho: Hello @etpinard, I switched to Yarn because Falcon uses Yarn. (You will notice this is a recurring pattern.) So, you would only want Yarn if there are clear upsides, but you also list one clear upside to Yarn. So, would you the program to use Yarn or NPM? I am happy to do it either way; let me know!

@n-riesco 
> BTW, Dockerfile and deployment/Dockerfile are still using npm.

On the off chance it's not just a tiny oversight, `./Dockerfile` does not exist:
```
$ find . -name Dockerfile
./deployment/Dockerfile
$
```

However, I would be happy to amend the existing Dockerfile.

package.json
JamesCropcho: @n-riesco I am inclined to keep cross-env in `dependencies{}` but I could easily move it to `devDependencies{}` if you'd like.

The thing is, this software is going to be executed by end-users in the command line (if only as a subprocess in a Pip package, or maybe more directly) on Windows and OS X environments. I worry about consistency across all environments with regards to environment variables.

@n-riesco et al, what do you think?

package.json
JamesCropcho: @n-riesco Please excuse this error or mine! I am sure I copied this from Falcon.

@etpinard @jackparmer @chriddyp Would you like me to put one of my email address in here? I suspect you'd prefer something which looks like `support@plotly.com`, and I probably would, too, but let me know.

package.json
JamesCropcho: @n-riesco I'm probably missing something silly here, but isn't `AppImage` on line 28?

.circleci/config.yml
etpinard: > but you also list one clear upside to Yarn.

The upside being a package lock file? In that case, that's not an upside, npm has one now. Voting for `npm` here.

.circleci/config.yml
JamesCropcho: @n-riesco It appears to me that "electron-pack-and-release" is running: https://circleci.com/gh/plotly/image-exporter/tree/test-builds-across-travis-circle-appveyor

(See #1124 for example.)

Please pardon, am I missing something simple?

package.json
chriddyp: you can put `chris@plot.ly` if you want

.circleci/config.yml
n-riesco: I have no preference.

package.json
n-riesco: @JamesCropcho That's not my understanding. We want to distribute `image-exporter` as an electron app, so that end-users don't have to install `node`.

If we include `cross-env` in `dependecies`, then `cross-env` will be included in the electron app (and the electron app doesn't use it).

.circleci/config.yml
n-riesco: @JamesCropcho For some reason, github doesn't show the build (as seen on the above screenshot). Since circleCI runs the job, then this isn't an issue.

.circleci/config.yml
JamesCropcho: Okay @n-riesco, I believe you'll like it now: https://circleci.com/gh/plotly/image-exporter/1172#artifacts/containers/0. What do you think?

.circleci/config.yml
JamesCropcho: Okay, @n-riesco, how about now?: https://circleci.com/gh/plotly/image-exporter/1175#config/containers/0

.circleci/config.yml
JamesCropcho: @n-riesco @etpinard Oh my, I am embarrassed at such a silly mistake!

Okay, I have updated `config.yml` to use the recommended key for Yarn (if/when we revert to NPM, I'll use the NPM key instead).

What do you think?

.circleci/config.yml
JamesCropcho: @n-riesco good call. Okay, I have added these caches: https://circleci.com/gh/plotly/image-exporter/1201#config/containers/0. I ran the builds twice after this change, to ensure it worked with and without a primed cache—all good. What do you think

JamesCropcho:
@n-riesco @etpinard 

> we need to disable asar

Done.

> getExportedArgs should use slice(1) instead of slice(2) when the executable from the installer is used.

Done.

appveyor.yml
JamesCropcho: @n-riesco  I am unsure why I chose those paths, however I feel most confident copying from Electron-Builder, so I have replaced my cache entries with theirs.

AppVeyor has not been running builds for the past fifteen minutes, so I am unable to verify a lack of errors at present.

n-riesco:
@JamesCropcho solving the issue with `getExportedArgs` requires a bit of work.

https://github.com/plotly/image-exporter/pull/63/commits/fea522b1330a34ee43a434edb3a470dfd0c1aa6b gets the electron app working, but breaks the "node" app.

@etpinard How about adding another file in `bin` to launch the electron app (and updating `package.json:main` accordingly)?

.circleci/config.yml
JamesCropcho: @etpinard @n-riesco I have reverted to NPM.

package.json
JamesCropcho: I have moved cross-env to `devDependencies`.

package.json
JamesCropcho: I have done so.

package.json
JamesCropcho: I have done this.

.circleci/config.yml
JamesCropcho: You are correct, @n-riesco . I have changed the CircleCI caching strategy as per your recommendations, and tested with fresh and primed caches.

JamesCropcho:
To my knowledge, no issues raised are presently unaddressed by me, nor is "the ball in my court" regarding any discussion subthreads on this pull request.

package.json
```diff
@@ -43,6 +60,7 @@
     "plotly"
   ],
   "dependencies": {
+    "ajv": "^6.4.0",
```
etpinard: what is this?

.circleci/config.yml
etpinard: I prefer locking down the image versions. I don't think this project is likely to break from one minor release of node.js to another.

JamesCropcho: Okay, I have made this change. How's it look?

package.json
JamesCropcho: @etpinard Adding this dependency resolved this error: https://github.com/epoberezkin/ajv/issues/708. If you like to learn more, you might start at https://github.com/npm/npm/issues/19877. I admit I have only cursorily glanced at it. More importantly, I got the above AJV error, and I fixed it by adding this explicit dependency (it was presumably already an implicit dependency).

Hope that explains things. Let me know if you have additional questions or thoughts about it.

n-riesco: @JamesCropcho but we don't use `ajv-keyword`, do we?

JamesCropcho: > I forgot about Étienne's comment: we should remove `ajv` from the deps.

Okay, I have done so,

etpinard:
Looking good except for that `docker-push` (intermittent?) failure. Anyone knows what's that about?

n-riesco:
@etpinard @JamesCropcho How about merging the `docker-push` and `docker-build` into one job? Then, there wouldn't be a chance of getting that `missing image` error.

etpinard:
> @etpinard @JamesCropcho How about merging docker-push and docker-build into one job? Then, there wouldn't be a chance of getting that missing image error.

It would be best to double-check with the docker-push step author @scjody before making a change like this.

JamesCropcho:
> How about merging docker-push and docker-build into one job?

Okay, I have done this.

JamesCropcho:
@etpinard @n-riesco 
> It would be best to double-check with the docker-push step author @scjody before making a change like this.

Just saw this. The commit may be easily reverted upon request.

scjody:
The reason `docker-push` and `docker-build` are separate steps are that this significantly cuts down test time by allowing `docker-build` to run alongside the other tests.

Has anyone looked into the "missing image" error? It might be worth talking to CircleCI support if it happens consistently.

Also while I'm here, I know @etpinard isn't big on rebases but I strongly suggest cleaning up the commit history of this PR before merging. It's up to you folks though.

etpinard:
> Also while I'm here, I know @etpinard isn't big on rebases but I strongly suggest cleaning up the commit history of this PR before merging. It's up to you folks though.

Yes, we should absolutely rebase these commits here.

n-riesco:
> Also while I'm here, I know @etpinard isn't big on rebases but I strongly suggest cleaning up the commit history of this PR before merging. It's up to you folks though.

Squash and merge?

JamesCropcho:
@scjody @etpinard @n-riesco 
> Squash and merge?

I too believe an automated squash-and-merge would be the local best-of-all-worlds.

I am not taking additional action with regards to Docker workflow unless instructed, or asked questions about it.

scjody:
My preference would be to rebase this down to 10 or so well thought out and understandable commits,  which will increase the usefulness of our git history (especially on existing files) vs. squashing.

However I realize that may be a significant amount of work and I'm not strongly against squashing.

etpinard:
Repost https://github.com/plotly/image-exporter/pull/63#issuecomment-380603152

> Has anyone looked into the "missing image" error? It might be worth talking to CircleCI support if it happens consistently.

JamesCropcho:
@etpinard 

> Has anyone looked into the "missing image" error? It might be worth talking to CircleCI support if it happens consistently.

I personally have not. I can talk to CircleCI support about it upon request, although I don't believe I would be best-qualified to do so. Also, I would be loathe to make this task a blocking issue for pull request acceptance.

etpinard:
Ok. Having talking to @scjody , merging docker-build and docker-push into one job isn't that big of a deal for now. Before the merger our total CI time was around 6 minutes, after the merged it is more like 8 minutes. Those two extra minutes is hardly worth time and effort at this stage. 

Issue logged in -> https://github.com/plotly/image-exporter/issues/68

etpinard:
Squashing and merging. Great job @JamesCropcho with special thanks to @n-riesco for the help!

JamesCropcho:
Thank you the kind words, @etpinard! I'm very happy to see this branch being merged into `master`.

PR #71, opened 2018-04-16
Bump vulnerable dev deps (take 2)

etpinard:
Supesedes https://github.com/plotly/image-exporter/pull/69 after I figured out the best way to _clear_ the CI `node_modules` cache (see https://github.com/plotly/image-exporter/pull/69 if that ever happens to you).

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -57,7 +57,7 @@ jobs:
       - run:
           name: List deps
           command: |
-            npm list
+            npm list || true
             pdftops -v
       - save_cache:
           key: v2-npm-{{ checksum "package-lock.json" }}
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -44,9 +44,9 @@ jobs:
           paths:
             - ".git"
       - restore_cache:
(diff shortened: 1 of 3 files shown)


etpinard:
merging this in.

etpinard:
Looks like this worked. That

image


warning is gone.

n-riesco:
So... the failure was just a cache corruption (because of the mismatch between `package.json` and `package-json.json`)?

etpinard:
Ah I don't think so. Where does CircleCi that into account the `package.json`?

JamesCropcho:
@etpinard @n-riesco 

> @JamesCropcho have you ever ran into issues like this one?

No (thankfully!).

> @etpinard I suspect the issue is that we are using the same keys to save caches for Node v6 and v8.

I too believe there's a high probability. Regardless, best-case scenario would be that this is a bad practice that would eventually cause some problem.

> Where does CircleCi take into account the package.json?

Off the top of my head, via `npm install` and, though a degree of separation (derivation)  `checksum "package-lock.json"`.

Issue #61, opened 2018-03-29
Intermittent errors in circle CI while running integration tests


tarzzz:
The integration tests seem to pass on local, but are very volatile on circle CI and fail often.
The situation has deteriorated with adding a new component (hence a new electron window) in: https://github.com/plotly/image-exporter/pull/60

cc: @etpinard ..

etpinard:
thanks for writing this down :writing_hand:

tarzzz:
The integration tests have been disabled in #60 for now.. !!

etpinard:
Followup on https://github.com/plotly/image-exporter/issues/61#issuecomment-377271980

Integration tests have been brought back in https://github.com/plotly/image-exporter/commit/433d619414c92f0c612d399cb72ca9a7b22bc5a0

etpinard:
PR https://github.com/plotly/image-exporter/pull/76 put `npm run test:integration` is 5x retry loop which should help make out CI runs be less intermittent.

Keeping this issue open, as #76 is far from an ideal fix to this issue.

etpinard:
Closing (for now).

PR #76, opened 2018-04-17
Run integration tests in retry loop

etpinard:
Similar to what was done in https://github.com/plotly/plotly.js/pull/2373, as a workaround those annoying intermittent test timeouts.


etpinard:
Related https://github.com/plotly/image-exporter/issues/61

etpinard:
Merging. This should help keep us sane while we search for a better solution for #61

PR #78, opened 2018-04-18
A few tweaks before 1st release

etpinard:
Mostly taking a few useful bits from https://github.com/plotly/image-exporter/pull/46 and polishing up a few things here and there.

cc @JamesCropcho


package.json
```diff
@@ -1,5 +1,5 @@
 {
-  "name": "@plotly/image-exporter",
+  "name": "plotly-image-exporter",
```
etpinard: For some reason `npm run pack` fails with `name: @plotly/image-exporter` (perhaps related to https://github.com/electron-userland/electron-builder/issues/911), so I changed it to `plotly-image-exporter`. That means we'll publish this project to npm as `plotly-image-exporter`. I hope no one minds.

JamesCropcho:
It's a potpourri, so it's hard to definitively give the stamp of approval, but yeah, given my perusal of all code changes, I see nothing that makes me worry.

package.json
etpinard: So to make things clear, 

- `plotly-image-exporter` is the name of the npm package to be published off this repo
- `plotly-graph-exporter` is the name of standalone app to be released in this repo

Rationale: 

Code in this repo is made to export more than just plotly graphs. For example, Dashboards, Dash reports and others things can now be exported using the image server. At the moment, we only create a standalone app to export plotly graphs. This standalone app is set to be used in the Python and R libraries. 

Down the road, if we want to create standalone apps for exporting Dash reports, I think we should name it something like `plotly-dash-report-exporter`. 

So, in brief, I found _image_ general enough to cover all things that can be exported with this tool and _graph_ precise enough to describe what the standalone app does.

I hope @jackparmer @chriddyp @cldougl @alexcjohnson @n-riesco @bpostlethwaite @tarzzz @JamesCropcho @jmmease @cpsievert agree.

n-riesco: Since the electron app only has one entry point (`package.json:main`), how about having a single app named `plotly-exporter` that takes as first argument `server`, `graph`, `dash-report`, ...?

etpinard: Interesting suggestion @n-riesco . I like it. Thanks! 

I'm not sure we _need_ it for the first release though. I don't think dash reports and dashboard export CLI runners are a huge priority for us in the few months (I could be wrong here).

etpinard: Augmenting on @n-riesco 's https://github.com/plotly/image-exporter/pull/78#discussion_r182646621, maybe we could expose that single entry point as `plotly-export` (instead of `plotly-exporter`), so that we get very clean sounding CLI commands like:

```
$ plotly-export graph first.json second.json https://plot.ly/~empet/10 --format pdf

$ plotly-export dash-report https://my-dash-report.com

$ plotly-export serve --port 9090
```

thoughts?

etpinard:
Ok. I'm going to merge this. The plotly-image-exporter -> orca rename effort will be part of a different PR. Follow https://github.com/plotly/image-exporter/issues/21 for more on that topic.

PR #82, opened 2018-04-25
Rename Image-Exporter to Orca

JamesCropcho:

PR #83, opened 2018-04-25
Rename Docker Image to Orca

JamesCropcho:


JamesCropcho:
@etpinard 
> I'm confused. What is the purpose of PR #83?

I am happy to address this question. The purpose of this pull request is to enable builds, after/if we rename the Docker Image to "orca" from "plotly-exporter."

Needless to say, I am in favor of this change, though as always (with regards to Plotly) I view this opinion as but one among a chorus.

> This PR (#82) has all its commits.

This is not quite correct. One may examine the changes on this branch which are not present on #82 here: https://github.com/plotly/orca/pull/83/files.

> Moreover, looks like "docker-build-and-push" is failing.

Yeah, "docker-build-and-push" will indeed fail on this branch until/if the Docker Image is renamed to "Orca."

JamesCropcho:
@etpinard 
> Oh I see #83 wants to merge commits from remove-orca into rename-docker-image-to-orca which is a PR-less branch.

In fact, rename-docker-image-to-orca is not a PR-less branch. Rather, its PR is at https://github.com/plotly/orca/pull/82. Each of the two branches has an associated PR.

> Here's what you should do instead: Create a PR that renames plotly-exporter -> orca everywhere except things related to quay.io and makes the tests pass, so that we can merge that thing ASAP.

The above is essentially what I have already done. Except, it is outside of my control but, although all tests pass (i.e. `npm run test`), the "docker-build-and-push" step will fail. This is because the "smoke test" cannot find the "image-exporter" image which possesses the `$CIRCLE_SHA1` tag, because all new commits are pushed to the "orca" repository, rather than the "image-exporter" repository (see "last modified": https://quay.io/repository/plotly/image-exporter?tab=tags).

> Next,  branch off the first PR and create another PR that make all the quay.io  changes that are not so important for the first release and thus can  wait before being merged.

I have also essentially done this. I'm am writing this message, within the PR you describe above.

===

What do you think of all this? What should my next actions be, all?

etpinard:
Ok, well arguing about this is a royal waste of time. Looks like you think your way is better. Fine. But, please just get the tests to pass one of those PRs w/o having to update quay.io things?

JamesCropcho:
@etpinard 
> Ok, well arguing about this is a royal waste of time. Looks like you think your way is better. Fine. But, please just get the tests to pass off one of those PRs w/o having to update quay.io things?

I am fine to do things your way, too, I promise!

What I was explaining in my most recent message on this pull request, is that since we changed the Git repository's name, the list of Docker Image tags at https://quay.io/repository/plotly/image-exporter?tab=tags seems to longer be updating. Therefore, it seems any changes pushed to the Orca codebase do not presently make their way to the Docker Image repository on Quay.

Therefore, it seems (kindly correct me if I'm mistaken) that "docker-build-and-push" cannot succeed with the present Quay configuration.

You are interested in renaming this image export API smartly and rapidly, and so am I, so that puts us on the same team here, Étienne. I have never used Quay before, so I am no expert: if there is some way to get "docker-build-and-push" to succeed without modifying the Quay configuration, let me know! I would get to it within the hour.

====

UPDATE: I may (?) have found a way. I am exploring this now. I am still open to hearing any thoughts which anyone may have, regarding.

JamesCropcho:
@etpinard I appear to have fixed the build on #82, albeit by thinking way outside the box, relative to the task-at-hand: https://github.com/plotly/orca/commit/9aefae340488aac8d6767bc60e23c932c50f99d6

deployment/Dockerfile
```diff
@@ -5,7 +5,7 @@ FROM ubuntu:xenial
 # From: https://github.com/nodejs/docker-node/blob/master/6.11/Dockerfile
 
 RUN apt-get update && apt-get install -y --no-install-recommends \
-        gnupg curl ca-certificates xz-utils wget libgconf-2-4 \
+        gnupg curl ca-certificates xz-utils wget libgtk2.0-0 libgconf-2-4 \
```
etpinard: So if I understand correctly, `docker-build-and-push` has been broken since we changed the repo's name to `orca`. This [dummy](https://github.com/plotly/orca/compare/dummy) branch [seems](https://circleci.com/workflow-run/8712231e-8cde-4da6-a767-0647a011db94) to confirm the issue. 

Now, why does adding `libgtk2.0` fix this issue?

scjody: I'm wondering this too. (As a future note, it would be great to put an explanation in the commit comment.)

JamesCropcho:
@scjody 
> @JamesCropcho Can you please produce a single PR against master that I can review? I don't think it makes sense to review this PR in its current form since it depends on (and even reverts) a bunch of stuff in #82.

Totally fair, Jody. I have "dismissed [your] review" for now; I could see why this was confusing. I suspect it would make more sense to review this pull request once/if #82 is merged.

etpinard:
Outdated. Closing.

PR #94, opened 2018-06-01
Use R's new orca() function

cpsievert:


README.md
```diff
@@ -87,11 +87,7 @@ To call `orca` from an R script:
 library(plotly)
 
 p <- plot_ly(x = 1:10, y = 1:10, color = 1:10)
-b <- plotly_build(p)$x[c("data", "layout")]
-json <- plotly:::to_JSON(b)
-
-cmd <- sprintf("orca graph '%s' -o r-export-test.png", json)
-system(cmd)
+orca(p, "plot.svg")
```
etpinard: NICE! Maybe you could list which version of the R library users need in order to use `orca`?

etpinard:
Merging. The python section doesn't state which version its `orca` wrapper got introduced, so let's merge this in.

PR #98, opened 2018-06-26
Add 'table' safeMode specs

etpinard:
I missed out on `table` traces in https://github.com/plotly/orca/pull/55, this PR adds them in.

cc @nicolaskruchten

### DIFF
--- a/src/component/plotly-graph/parse.js
+++ b/src/component/plotly-graph/parse.js
@@ -148,23 +148,27 @@ function findMaxArrayLength (cont) {
     }
   })
 
-  let l = Math.max(0, ...lengths)
-
-  if (cont.type === 'table' && isPlainObj(cont.cells)) {
-    l = Math.max(l, findMaxArrayLength(cont.cells))
-  }
-  return l
+  return Math.max(0, ...lengths)
 }
 
 function estimateDataLength (trace) {
+  const topLevel = findMaxArrayLength(trace)
+  let dimLevel = 0
+  let cellLevel = 0
+
   // special case for e.g. parcoords and splom traces
   if (Array.isArray(trace.dimensions)) {
-    return trace.dimensions
+    dimLevel = trace.dimensions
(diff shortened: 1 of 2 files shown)


src/component/plotly-graph/parse.js
```diff
@@ -148,7 +148,12 @@ function findMaxArrayLength (cont) {
     }
   })
 
-  return Math.max(0, ...lengths)
+  let l = Math.max(0, ...lengths)
+
+  if (cont.type === 'table' && isPlainObj(cont.cells)) {
+    l = Math.max(l, findMaxArrayLength(cont.cells))
+  }
```
alexcjohnson: `table` doesn't have any arrays in the top level container, does it? I might have put this in `estimateDataLength` then, alongside the similar logic for `dimensions`. Also this way we'd keep looking to recurse, `trace.cells.cells.cells`...

That said, the fact that this combines with the top level instead of replacing it is nice. Just imagining malicious users, the simple test `if (Array.isArray(trace.dimensions))` as it is with an immediate return could be used on a trace that doesn't use `dimensions` to mask the fact that the real data arrays are huge.

So we could add a type qualifier to `dimensions`, but perhaps the most robust solution (and future-proof as we add more trace types so we don't need to worry too much about updating the logic here) would be to look at the top level AND in `cells` AND in `dimensions` regardless of trace type, and find the max length of everything. I don't think anyone is going to be bothered by the fact that their useless length-1e8 `cells.junk` array in a scatter trace got rejected 😈

etpinard: > would be to look at the top level AND in cells AND in dimensions regardless of trace type, and find the max length of everything.

I like this solution a lot. Done in https://github.com/plotly/orca/pull/98/commits/2b5cf94ab214c52f988303878d46fe894e0bc80b

alexcjohnson:
Nice job! 💃

PR #105, opened 2018-07-30
Release 1.0.0-onprem

tarzzz:


etpinard:
thanks :dancer: !

etpinard:
The `v1.0.0-onprem` container is now up on https://quay.io/repository/plotly/image-exporter?tab=tags :tada: 

@tarzzz I'm guessing you're leaving this PR open in case things need to be patched for the next On-Prem?

tarzzz:
> I'm guessing you're leaving this PR open in case things need to be patched for the next On-Prem?

We can merge it after this on-prem release..

Issue #110, opened 2018-08-06
Orca hanging on large json files


sdrap:
I have a dataframe with 20 years of daily data. iplot can process any plots.

However, I can only use orca when I slice the dataframe for less than 4 years of data.
It fails in the notebook as well as in command line from a json dumped file with the following text.

```
A JavaScript error occurred in the main process
Uncaught Exception:
TypeError: path must be a string or Buffer
    at Object.fs.mkdirSync (fs.js:891:18)
    at main (/usr/local/lib/node_modules/orca/bin/graph.js:105:8)
    at Object.<anonymous> (/usr/local/lib/node_modules/orca/bin/orca_electron.js:73:25)
    at Object.<anonymous> (/usr/local/lib/node_modules/orca/bin/orca_electron.js:99:3)
    at Module._compile (module.js:569:30)
    at Object.Module._extensions..js (module.js:580:10)
    at Module.load (module.js:503:32)
    at tryModuleLoad (module.js:466:12)
    at Function.Module._load (module.js:458:3)
    at loadApplicationPackage (/usr/local/lib/node_modules/electron/dist/resources/default_app.asar/main.js:287:12)

```
The JSON files are 250Kb large for the 20 years of data.
[data.zip](https://github.com/plotly/orca/files/2264157/data.zip)

etpinard:
I suspect you're calling orca as:

```js
orca graph data.zip
```

which won't work, as we currently only accept path/to/json/files, url/to/json/files or json strings.

Is this indeed the case?

sdrap:
No I uploaded as a zip because github doesn't accept plain json. I run it as follows

```
orca graph data.json -o test.png -debug
```

or alternatively

```
cat data.json | orca graph > test.png
```
which just hang.

I don't know exactly what is the problem.

etpinard:
I think something is up with your data encoding:

```
var s = fs.readFileSync('./data.json');
typeof JSON.parse(s)
// => return 'string'
```

Orca expect the result of `JSON.parse` to an object.

sdrap:
I am puzzled a bit with JSON. I dumped in the file using

```
A = df['S'].iplot(**kwargs)
file = 'data.json'
with open(file, 'w') as outfile:
    json.dump(json.dumps(A, cls=plotly.utils.PlotlyJSONEncoder), outfile)
```

Strangely enough it starts with `"{` and put a backslash in front of every double quote.
I edited the file removing these characters in a working json that I can process through cat.

[...]

PR #112, opened 2018-08-09
Add --graph-only option for the orca serve entry point

jonmmease:
## Overview
Added a `--graph-only` boolean command line option to the `orca serve` entry point. When this option is present, only the `plotly-graph` component is run (not thumbnail, dash, etc.).

When graph conversion is the only service needed, this option reduces the number of electron processes, the memory usage, and startup time.

## Past Discussion
See https://github.com/plotly/orca/issues/110

## Resource usage comparison
Here's a comparison of the baseline resource usage on OSX

### Without `--graph-only` (legacy behavior)
screen shot 2018-08-09 at 8 15 23 am

There are 8 processes running, consuming ~390MB of memory.

Based on my testing from Python, from launching the server to receiving the first image conversion takes about 2.2 seconds.

### With `--graph-only
screen shot 2018-08-09 at 8 15 59 am

There are 3 processes running, consuming ~120MB of memory.

Based on my testing from Python, from launching the server to receiving the first image conversion takes about 1.7 seconds.

### Resource Summary
If all you need is graph conversion, this flag saves 5 processes, 270MB of memory, and a half second of startup time.

## Testing
I added a new integration testing file at `test/integration/orca_serve_graph-only_test.js`.  This file is based on a subset of the testing in `test/integration/orca_serve_test.js`. But it checks to make sure there is only 1 window running, that the graph-component is running, and that the thumbnail component is not running.

### DIFF
--- a/bin/serve.js
+++ b/bin/serve.js
@@ -28,6 +28,11 @@ const OPTS_META = [].concat([{
   alias: ['windowMaxNumber', 'maxNumberOfWindows'],
   description: 'Sets maximum number of browser windows the server can keep open at a given time.'
 }, {
+  name: 'graph-only',
+  type: 'boolean',
+  alias: ['graphOnly'],
+  description: 'Launches only the graph component (not thumbnails, dash, etc.) to save memory and reduce the number of processes'
+}, {
   name: 'quiet',
   type: 'boolean',
   description: 'Suppress all logging info.'
@@ -72,15 +77,14 @@ function main (args) {
       console.log(`Spinning up server with pid: ${process.pid}`)
     }
 
-    app = orca.serve({
-      port: opts.port,
-      maxNumberOfWindows: opts.maxNumberOfWindows,
-      debug: opts.debug,
-      component: [{
-        name: 'plotly-graph',
-        route: '/',
-        options: plotlyJsOpts
-      }, {
+    let component = [{
+      name: 'plotly-graph',
+      route: '/',
+      options: plotlyJsOpts
+    }]
+
+    if (!opts.graphOnly) {
+      component.push({
         name: 'plotly-dashboard',
         route: '/dashboard'
       }, {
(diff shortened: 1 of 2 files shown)


bin/serve.js
```diff
@@ -27,6 +27,11 @@ const OPTS_META = [].concat([{
   type: 'string',
   alias: ['windowMaxNumber', 'maxNumberOfWindows'],
   description: 'Sets maximum number of browser windows the server can keep open at a given time.'
+}, {
+  name: 'graph-only',
+  type: 'boolean',
+  alias: ['graphOnly'],
+  description: 'Launches only the graph component (not thumbnails, dash, etc.) to save memory and reduce the number of processes'
```
etpinard: :cow2: `.` at the end.

etpinard:
Brilliant PR @jonmmease ! Thanks very much!

Now, that `docker-build-and-push` job is failing again, and but the looks of it it never ran successfully off one of your branches. 

Digging deeper, the environment variable `$DOCKER_USER` and `$DOCKER_PASS` are blank on runs off your branches whereas they are correctly filled otherwise. 

So, I think we should probably disable `docker-build-and-push` off forks. I hope there's a way to do that in the circle config. I doubt we want to pass environment variable to all open-source contributors. Alternatively, @jonmmease you could start branching off the official repo (I invited you to join as collaborator) instead of off your fork.

jonmmease:
Sure thing!

Yeah, it would be nice to disable the `docker-build-and-push` CI job for forks if there's a way. But in the meantime I'll go ahead and move my two current PRs to the main repo and see if that takes care of it.

jonmmease:
Replaced by https://github.com/plotly/orca/pull/114

bin/serve.js
jonmmease: Done in #114 🙂

PR #114, opened 2018-08-09
Server graph only

jonmmease:
This PR supersedes #112 due to CI permissions issues. 

## Overview
Added a `--graph-only` boolean command line option to the `orca serve` entry point. When this option is present, only the `plotly-graph` component is run (not thumbnail, dash, etc.).

When graph conversion is the only service needed, this option reduces the number of electron processes, the memory usage, and startup time.

## Past Discussion
See https://github.com/plotly/orca/issues/110

## Resource usage comparison
Here's a comparison of the baseline resource usage on OSX

### Without `--graph-only` (legacy behavior)
screen shot 2018-08-09 at 8 15 23 am

There are 8 processes running, consuming ~390MB of memory.

Based on my testing from Python, from launching the server to receiving the first image conversion takes about 2.2 seconds.

### With `--graph-only
screen shot 2018-08-09 at 8 15 59 am

There are 3 processes running, consuming ~120MB of memory.

Based on my testing from Python, from launching the server to receiving the first image conversion takes about 1.7 seconds.

### Resource Summary
If all you need is graph conversion, this flag saves 5 processes, 270MB of memory, and a half second of startup time.

## Testing
I added a new integration testing file at `test/integration/orca_serve_graph-only_test.js`.  This file is based on a subset of the testing in `test/integration/orca_serve_test.js`. But it checks to make sure there is only 1 window running, that the graph-component is running, and that the thumbnail component is not running.

### DIFF
--- a/bin/serve.js
+++ b/bin/serve.js
@@ -28,6 +28,11 @@ const OPTS_META = [].concat([{
   alias: ['windowMaxNumber', 'maxNumberOfWindows'],
   description: 'Sets maximum number of browser windows the server can keep open at a given time.'
 }, {
+  name: 'graph-only',
+  type: 'boolean',
+  alias: ['graphOnly'],
+  description: 'Launches only the graph component (not thumbnails, dash, etc.) to save memory and reduce the number of processes'
+}, {
   name: 'quiet',
   type: 'boolean',
   description: 'Suppress all logging info.'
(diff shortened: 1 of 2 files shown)


etpinard:
Beautiful, merge away :dancer:

Issue #120, opened 2018-08-15
Orca Error -  sh: orca: command not found


Transmystical:

etpinard:
Is `plotly-orca` part of your `conda list` output?

Transmystical:
Thank you for your response! Yes, it is:

$ conda list
plotly-orca               1.1.0                         1    plotly


Plus, I can do this …. 

$ orca --help
Plotly's image-exporting utilities

  Usage: orca [--version] [--help] <command> [<args>]

  Available commands:
  - graph [or plotly-graph, plotly_graph] 


However, my error has now changed when I run the following command: 

json <- plotly:::to_JSON(p)
cmd <- sprintf("orca graph '%s' -o r-export-test.pdf", json)
system(cmd) 

/usr/local/bin/orca: line 2: /Applications/orca.app/Contents/MacOS/orca: No such file or directory
/usr/local/bin/orca: line 2: exec: /Applications/orca.app/Contents/MacOS/orca: cannot execute: No such file or directory


So, I did the following: 

$ cd /usr/local/bin
$ cat orca 
#!/bin/bash
exec "/Applications/orca.app/Contents/MacOS/orca" "$@“


I think the 2nd line it cannot execute for some reason. I went to my /Applications and there is no Applications/orca.app 
Is there supposed to be? 


Many thanks!

etpinard:
I think this might a case of conflicting installations. Have you tried installing orca by other means than `conda`?

Michelebradley:
Getting the same error as @Transmystical after using the system(cmd) prompt

Downloaded the standalone Orca binaries for Mac

CavinWard:
Just wanted to ask if this issue has been resolved, and if so what the current fix is. I tried exporting a plot using orca in R (R 3.5.1 and plotly 4.8.0) today (9/9/2018) after installing orca using conda . Running `orca(p, "test.png")` gave the following error 
> Error: The orca command-line utility is required to use the `orca()` function
Follow the installation instructions here -- https://github.com/plotly/orca#installation

I can run `help(orca)` with no errors

etpinard:
> I can run help(orca) with no errors

That gives no errors, but does it output anything?

rickyars:
I'm having the same problem as @CavinWard. I just installed `orca` using on my Mac using this command:
```
conda install -c plotly plotly-orca
```
I can see orca:
```
$ which orca
/anaconda3/bin/orca
$ orca --version
1.1.1
```
[...]

Issue #111, opened 2018-08-08
Permission denied downloading orca?


mpowell12:
Command failed: cp /Applications/orca.app/Contents/Resources/app/bin/orca.sh /usr/local/bin/orca
cp: /usr/local/bin/orca: Permission denied

error message for mac

etpinard:
`/usr/local/bin/orca` is usually a protected directory.

You could try to add `sudo` at the start of that command (if you have admin rights on your machine of course), or cp `orca` to a directory in your `$PATH` that you have access to.

mpowell12:
this worked! thank you

jonmmease:
See PR https://github.com/plotly/orca/pull/122 for a proposed solution to avoid this permissions error.

Issue #92, opened 2018-05-17
Make "easy install" shell scripts for Mac and Windows


etpinard:
... to hopefully turn [Mac](https://github.com/plotly/orca#mac-os) and [Windows](https://github.com/plotly/orca#windows) checklist into a single command.

cc @n-riesco

n-riesco:
It'd be good to have a permanent link to the latest installer for each platform. Something like [Falcon's](https://plot.ly/free-sql-client-download/). It'd make the install scripts easier to write.

etpinard:
good call!

PR #122, opened 2018-08-25
Improvements to "Standalone binaries" installation processes

jonmmease:
## Overview
In preparation for the coming integration of orca into plotly.py, I reviewed and tested the "Standalone binaries" installation instructions for all three operating systems.  This PR contains the updates from this review.

In all cases, the goal is to make sure the user ends up with an executable named `orca` on the permanent system `PATH`.  This way the default plotly.py orca search logic will find it automatically.

I believe this is the last set of orca changes that we need before the plotly.py release. The Windows and Linux changes are documentation only.  The Mac OS changes include both documentation and installer updates.

I think this is the last set of changes in orca that we need before the plotly.py release.  I was picturing that we would release these changes as version 1.1.1 as soon as this PR is merged.

## Windows
The Windows installation instructions have been updated to direct the user to add the orca executable to their `PATH` permanently. The prior instructions only added it to the `PATH` for the current Command Prompt session. 

The actual environment variable instructions (for each Windows version) are delegated to a nice article on computerhope.com (https://www.computerhope.com/issues/ch001647.htm)

## Linux
The Linux installation instructions have been updated to direct the user to create a symbolic link named `orca` that points to the AppImage.  This way the user ends up with something name `orca` on their path (not `orca-X.Y.Z-x86_64.AppImage`).

Another computerhope.com reference was added to explain the concept of the `PATH` on Linux (https://www.computerhope.com/issues/ch001647.htm), along with a reference with some basic AppImage information (https://appimage.org/).

## Mac OS
The logic to add orca to the system `PATH` has been refined as follows:
 - Don't make any changes if `orca` is already on the `PATH`.
 - Only copy `orca.sh` to `/usr/local/bin` if orca is installed as an application (not if it's installed through conda or npm). I think this will address the trouble folks are running into in https://github.com/plotly/orca/issues/120.
 - Perform the `orca.sh` copy with administrator privileges to avoid permission denied errors (See https://github.com/plotly/orca/issues/111). This is done using Apple Script, which launches a nice GUI password prompt.

The README instructions are also updated to reflect the new installation process.

### DIFF
--- a/README.md
+++ b/README.md
@@ -53,10 +53,23 @@ your operating system from the
 
 #### Windows
 
-- Extract the `window-release.zip` file
-- In the `release` folder, double-click on `orca Setup X.Y.Z`, this will create an orca icon on your Desktop
-- Right-click on the orca icon, then click on _Properties_ and copy the _Starts in_ field
-- In the command prompt, run `PATH %PATH%;<paste the "Starts in" field here>`
+- Extract the `windows-release.zip` file.
+- In the `release` folder, double-click on `orca Setup X.Y.Z`, this will create an orca icon on your Desktop.
+- Right-click on the orca icon and select _Properties_ from the context menu.
+- From the _Shortcut_ tab, copy the directory in the _Start in_ field.
+- Add this _Start in_ directory to you system `PATH` (see below).
+- Open a new Command Prompt and verify that the orca executable is available on your `PATH`.
+
+```
+> orca --help
+Plotly's image-exporting utilities
+
+  Usage: orca [--version] [--help] <command> [<args>]
+  ...
+```
+
+##### Windows References
+ - How to set the path and environment variables in Windows: https://www.computerhope.com/issues/ch000549.htm
 
 #### Linux
 
--- a/README.md
+++ b/README.md
@@ -93,7 +93,7 @@ Plotly's image-exporting utilities
 $ chmod +x orca-X.Y.Z-x86_64.AppImage
 ```
 
-- Create a symbolic link named `orca` somewhere on your `$PATH` that points
+- Create a symbolic link named `orca` somewhere on your `PATH` that points
 to the AppImage.
 
 ```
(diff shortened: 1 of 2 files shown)


bin/orca_electron.js
```diff
@@ -44,24 +44,54 @@ if (process.platform === 'darwin' && process.argv.length === 1) {
...
 
-  const source = path.join(__dirname, 'orca.sh')
-  const target = '/usr/local/bin/orca'
+  let showMessage = false;
 
-  if (!fs.existsSync(target)) {
-    try {
-      execSync(`cp ${source} ${target}`)
+  try {
+      execSync('which orca');
```
etpinard: Ha interesting. I didn't think this would lead to an exception.

README.md
```diff
@@ -53,10 +53,23 @@ your operating system from the
...
 #### Windows
 
-- Extract the `window-release.zip` file
-- In the `release` folder, double-click on `orca Setup X.Y.Z`, this will create an orca icon on your Desktop
-- Right-click on the orca icon, then click on _Properties_ and copy the _Starts in_ field
-- In the command prompt, run `PATH %PATH%;<paste the "Starts in" field here>`
+- Extract the `windows-release.zip` file.
+- In the `release` folder, double-click on `orca Setup X.Y.Z`, this will create an orca icon on your Desktop.
+- Right-click on the orca icon and select _Properties_ from the context menu.
+- From the _Shortcut_ tab, copy the directory in the _Start in_ field.
```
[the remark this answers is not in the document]
jonmmease: Correct, There's also a *Target* field that is the full path to the executable.

etpinard:
Thanks for doing this @jonmmease :tada: 

Now [`docker-build-and-push`](https://circleci.com/gh/plotly/orca/1998) is failing. Looks like `svn` can't resolve https://github.com/plotly/plotly.js.git/trunk/dist/extras/mathjax anymore. This probably has nothing to do with your PR.

jonmmease:
I just pushed some additional "Linux troubleshooting" instructions that cover the case of installing orca on Ubuntu Server (and they should at least be helpful to people working with other server distributions).  This requires adding a handful of additional system libraries plus Xvfb.

In the instructions I suggest creating a shell script that runs the orca AppImage executable using `xvfb-run`. In my testing this works really well, and this way plotly.py won't need any Xvfb specific logic.

See what you think!

README.md
```diff
@@ -44,25 +44,120 @@ your operating system from the
...
+```
+$ orca --help
+orca: error while loading shared libraries: libgtk-x11-2.0.so.0:
+cannot open shared object file: No such file or directory
+```
+
+On Ubuntu Server, these additional libraries can be installed with:
+
+```
+sudo apt-get install libgtk2.0-0 libxtst6 libxss1 libgconf-2-4 libnss3 libasound2
```
etpinard: Could you instead tell users to install `google-chrome` (like in our image-server [Dockerfile](https://github.com/plotly/orca/blob/4bd21148fe4a1f13ee783b20b26fbf8788186583/deployment/Dockerfile#L90-L94)? That and `libgtk2.0-0` / `libgconf-2-4` should be enough to run orca smoothly

etpinard:
> In the instructions I suggest creating a shell script that runs the orca AppImage executable using xvfb-run. In my testing this works really well, and this way plotly.py won't need any Xvfb specific logic.

We use `xvfb-run` 

https://github.com/plotly/orca/blob/4bd21148fe4a1f13ee783b20b26fbf8788186583/deployment/run_server#L29

in our image-server container, so yeah it should work w/o any issues.

README.md
etpinard: I would prefer if the installation instructions match something that we can test.

Pointing users to the what we use in our image-server container should achieve that (as we test it on every push and on prod). But, perhaps our Dockerfile is a little too Ubuntu-centric? Do you think many people with try to user orca in non-Ubuntu server environments?

jonmmease: Sounds good. Done in a87754d 

Let me know if you think of anything else!

etpinard:
Thanks very much @jonmmease !

:dancer:

jonmmease:
Sure thing! Hopefully this will help cut down on the number orca and plotly.py tickets 🙂

etpinard:
> Hopefully this will help cut down on the number orca and plotly.py tickets

Yep, this PR is a solid step forward, until we take a stab at https://github.com/plotly/orca/issues/92

jonmmease:
Now that we don't need to get https://github.com/plotly/orca/pull/125 merged before plotly.py 3.2.0 is released, how does it sound to merge this PR and release it as 1.1.1?  Would you have time to do the release by the end of the week? Thanks!

etpinard:
> Would you have time to do the release by the end of the week

Sure. I'll do it before lunch today.

Issue #124, opened 2018-08-29
Segmentation fault when NODE_OPTIONS environment variable set


jonmmease:
See Electron issue https://github.com/electron/electron/issues/12695

Prior to electron version 2.0.3., having the `NODE_OPTIONS` environment variable set causes electron applications to exit with a segmentation fault, and I can reproduce this behavior with orca.

This is particularly unfortunate for plotly.py because the [instructions](https://github.com/plotly/plotly.py) for installing the plotly.py JupyterLab extensions require setting `NODE_OPTIONS=--max-old-space-size=4096` to avoid out-of-memory errors.

See https://github.com/plotly/orca/issues/89 for @etpinard's notes on Electron 2.0.  It think I'll try out 2.0.3+ today and see if anything breaks. It would be nice to make this change before the plotly.py 3.2.0 release if it's not too large of an effort.

PR #125, opened 2018-08-29
Update electron version to 2.0.8 to address segmentation fault

jonmmease:
This PR updates the electron version from 1.8.4 to 2.0.8 to address the segmentation fault error discussed in #124.

I verified on MacOS that the segmentation fault no longer occurs when the NODE_OPTIONS environment variable is set.

My local tests all passed without an issue, now we'll see what the CI tests do.

### DIFF
--- a/README.md
+++ b/README.md
@@ -31,7 +31,7 @@ If you have Node.js installed (recommended v8.x), you can easily install Orca
 using npm as:
 
 ```
-$ npm install -g electron@1.8.4 orca
+$ npm install -g electron@2.0.8 orca
 ```
 
 which makes the `orca` executable available in your path.
--- a/README.md
+++ b/README.md
@@ -31,7 +31,7 @@ If you have Node.js installed (recommended v8.x), you can easily install Orca
 using npm as:
 
 ```
-$ npm install -g electron@2.0.8 orca
+$ npm install -g electron@2.0.9 orca
 ```
 
 which makes the `orca` executable available in your path.
--- a/deployment/README.md
+++ b/deployment/README.md
@@ -33,6 +33,12 @@ This is done using plotbot. The following commands provide help:
 ```
 @plotbot deploy how
 @plotbot run how
+
+# for example,
+@plotbot deploy to imageserver stage
+
+@ or
+@plotbot deploy to imageserver prod
 ```
 
 # Font Support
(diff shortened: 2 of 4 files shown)


etpinard:
Great. 

We'll need to test this on stage, to make sure the new electron behaves well enough for our image-server before merging this. 

@scjody what are some of the tests you ran last spring?

etpinard:
Thanks @scjody !

Unfortunately @jonmmease , I don't think I'll have time to get to this until at least Friday if not next Tuesday. 

If not too busy, perhaps @tarzzz could help us out?

jonmmease:
I totally understand @etpinard, this sounds a bit involved.  I'll focus on the rest of the plotly.py 3.2 release items and them make the call on whether to block on this or not. Thanks!

jonmmease:
@etpinard I'm in the process of adding a workaround to plotly.py that will simply unset the `NODE_OPTIONS` environment variable before calling orca. See https://github.com/plotly/plotly.py/pull/1140/.

This takes care of the segmentation fault problem when orca is used from a plotly.py session that was launched with the `NODE_OPTIONS` environment variable set. Now that I know that this workaround is possible, I don't think we need to block plotly.py 3.2.0 on this PR 🙂

etpinard:
> I don't think we need to block plotly.py 3.2.0 on this PR

Is this still the case @jonmmease ?

jonmmease:
Yes.  My workaround is working well. Thanks for following up!

etpinard:
> Yes. My workaround is working well. Thanks for following up!

Fantastic. I'll differ testing Electron 2.0 to after the plotly.js `v1.41.0` release (in hopefully about ~1 week).

tonkolviktor:
Hello,

+1

On raspberry pi 3 b+ with version 1.8.4 of electron and latest node [1] i get the following error [2].
The reason is that the server is never started, because "did-finish-load" event is never emitted in src/app/server/index.js. (I do not know why or why the page is not loaded!?)

After applying the changes manually in this PR it works for me, would be nice to have it in the master ;)

(BTW first i tried with electron 5.X, the server could start but the plot generation was just hanging without any apparent reason)

[1]
pi@raspberrypi:~ $ node -v
v10.15.3
pi@raspberrypi:~ $ npm -v
6.9.0

[2]
ValueError: 
For some reason plotly.py was unable to communicate with the
local orca server process, even though the server process seems to be running.

antoinerg:
Superseded and closed by https://github.com/plotly/orca/pull/266

Issue #41, opened 2017-12-08
Steadily increasing memory usage


scjody:
After a 2 hour run of all our mocks except `gl*` and `mapbox*` using `test/image-make_baseline.js` (351 mocks, taking about 55 seconds per run), against image-exporter running as an imageserver in a Docker container, memory usage grew steadily:

https://plot.ly/%7EJodyMcintyre/2209/
![new imageserver memory usage](https://user-images.githubusercontent.com/1136329/33743253-0c5987b0-db7a-11e7-841d-e7fba48b4219.png)

When the run was stopped, memory usage decreased a bit then leveled out.

Examination of `ps` results showed that two `electron` processes (probably the plot image and plot thumbnail processes) were responsible for the memory usage:

![screen shot 2017-12-07 at 18 00 40](https://user-images.githubusercontent.com/1136329/33743287-266a295c-db7a-11e7-9683-83c2bda5d08b.png)

A similar issue was observed in a 12 hour run of 3 imageservers in our staging environment:

![screen shot 2017-12-08 at 15 01 56](https://user-images.githubusercontent.com/1136329/33782987-ce5834de-dc28-11e7-9ffc-1c9e7b307d09.png)

As a workaround for this issue we could restart Electron after a reasonably large number of requests (e.g. 1,000). I'm going to look into this but a fix for the root cause is needed at some point.

@etpinard @monfera FYI

mag009:
I was able to reproduce the error feeding sequential request of a large json file ( 4.7M ) and every single time at 233 request the process crash. The larger the file is the quicker the process crash.

- it's not a resources issue as no message appear in dmesg nor a segfault. 
- I've tested with current version of electron and 2.0.9 same result.

 I've tested with the following file : success/ee151de8-d07e-4d19-a2b6-400a2292c609_200.json
from : [https://github.com/plotly/streambed/issues/9865#issuecomment-360823004](url)
`
 for i in {0..5000}; do curl -d @ee151de8-d07e-4d19-a2b6-400a2292c609_200.json http://localhost:9091/ ;echo $i;done
`

capture of orca's container when the 233rd requests happen: 

12544ae95ac6        eager_banach          149.46%             2.447GiB / 4GiB       61.18%              0B / 0B             0B / 303kB          153

error message : 

<--- Last few GCs --->

[...]

PR #128, opened 2018-09-11
add resources limit for proper scaling

mag009:
See plotly/streambed#9865 and plotly/streambed#11037

Reason for removing the --request-limit is because currently when we hit the 1000 requests the server exit but the container stays up monit handle the restart and when you do that the container is still healthy on the LB. So it is possible that client connect and hit a connection refused.

To avoid that I'm limiting the resources of the container cpu and memory in which if the app have a memory leak it will kill the container and making it unavailable to the LB and just spin a new container.

The annotation is required to scale-down, if a container spin-up on a node where kube-system is running so it tells it's okay to kill that node.

- [ ] issue with auto-balance doesn't work doesn't seem to down-scale node that should
- [ ] fined tuning the resources
- [ ] preemptive instances

tested with : 
ab -r -c 100 -n 100000 -p 86eac25f-4de9-4da2-82a9-0c7d28db1454_200.json http://10.128.0.17:9091/

### DIFF
--- a/deployment/kube/stage/frontend.yaml
+++ b/deployment/kube/stage/frontend.yaml
@@ -5,6 +5,8 @@
 apiVersion: extensions/v1beta1
 kind: Deployment
 metadata:
+  annotations:
+    cluster-autoscaler.kubernetes.io/safe-to-evict: "true"
   name: imageserver
   labels:
     app: imageserver
@@ -17,6 +19,15 @@ spec:
         tier: frontend
     spec:
       affinity:
+        podAffinity:
+          requiredDuringSchedulingIgnoredDuringExecution:
+            - labelSelector:
+                matchExpressions:
+                  - key: "app"
+                    operator: In
+                    values:
+                    - imageserver
+              topologyKey: "failure-domain.beta.kubernetes.io/zone"
         podAntiAffinity:
           requiredDuringSchedulingIgnoredDuringExecution:
             - labelSelector:
@@ -43,6 +54,13 @@ spec:
         ports:
         - name: http-server
           containerPort: 9091
+        resources:
+         limits:
+           cpu: 600m
+           memory: 1Gi
+         requests:
+           cpu: 400m
+           memory: 600Mi
         volumeMounts:
         - mountPath: "/usr/share/fonts/user"
           name: plotly-cloud-licensed-fonts
--- a/deployment/kube/stage/hpa.yaml
+++ b/deployment/kube/stage/hpa.yaml
@@ -14,5 +14,5 @@ spec:
   # Set this to 3x "min-nodes":
   minReplicas: 3
   # Set this to 3x "max-nodes":
-  maxReplicas: 3
+  maxReplicas: 6
   targetCPUUtilizationPercentage: 50
--- a/deployment/kube/stage/pdb.yaml
+++ b/deployment/kube/stage/pdb.yaml
@@ -1,0 +1,9 @@
+apiVersion: policy/v1beta1
+kind: PodDisruptionBudget
+metadata:
+  name: imageserver-pdb
+spec:
+  minAvailable: 2
+  selector:
+    matchLabels:
+      app: imageserver
--- a/deployment/run_server
+++ b/deployment/run_server
@@ -26,7 +26,7 @@ fi
 pkill Xvfb
 pkill node
 
-xvfb-run --auto-servernum --server-args '-screen 0 640x480x24' ./bin/orca.js serve --request-limit=1000 --safe-mode $PLOTLYJS_ARG $@ 1>/proc/1/fd/1 2>/proc/1/fd/2 &
+xvfb-run --auto-servernum --server-args '-screen 0 640x480x24' ./bin/orca.js serve --safe-mode $PLOTLYJS_ARG $@ 1>/dev/stdout 2>/dev/stderr &
 echo \$! > \$PIDFILE
 
 EOF


etpinard:
Thanks @mag009 !

I'll let @scjody review this thing (I don't much about the `deployment/` folder).

One thing to keep in mind, I'll like to upgrade the electron version we're using (see https://github.com/plotly/orca/pull/125) in the short-term. So, maybe you run your tests using the [`electron_2.0.8` orca image on quay,io](https://quay.io/repository/plotly/image-exporter?tab=tags)? That would be much appreciated. Thanks!

mag009:
I was able to stress test and so far so good. No crashes in 30 minutes for a total of 26k requests, at an avg rate  of 17 req/s
 
I'm using preempt vm's with shared cpu which can handle ~2req/s. 

I still have minor issue, i'm getting connection refused when it add a container so i probably need to adjust the health check.

Test below is performed with a 384K file 


Completed 10000 requests
Completed 20000 requests

Server Software:        
Server Hostname:        10.128.0.17
Server Port:            9091

Document Path:          /
Document Length:        Variable

Concurrency Level:      15
Time taken for tests:   1481.105 seconds
Complete requests:      26407
Failed requests:        0
Total transferred:      3235762787 bytes
Total body sent:        10379777012
HTML transferred:       3232488319 bytes
Requests per second:    17.83 [#/sec] (mean)
Time per request:       841.314 [ms] (mean)
Time per request:       56.088 [ms] (mean, across all concurrent requests)
Transfer rate:          2133.49 [Kbytes/sec] received
                        6843.88 kb/s sent
                        8977.37 kb/s total

Connection Times (ms)
              min  mean[+/-sd] median   max
Connect:        0    7  20.1      2     292
Processing:   137  833 1475.3    315    8102
Waiting:      133  822 1467.5    306    8032
Total:        140  840 1482.5    321    8180

Percentage of the requests served within a certain time (ms)
  50%    321
  66%    421
  75%    530
  80%    616
  90%   2638
  95%   5146
  98%   6062
  99%   6462
 100%   8180 (longest request)

etpinard:
> I was able to stress test and so far so good.

Nice!


> Test below is performed with a 384K file

Can you point us to that file? I might nice to run the tests on a collections of [plotly.js mocks](https://github.com/plotly/plotly.js/tree/master/test/image/mocks) or using ["real-life" image server requests](https://github.com/plotly/streambed/issues/9865#issuecomment-360823004).

Last spring, @scjody used this [thing](https://github.com/plotly/plotly.js/blob/imageserver-testing-hacks/test/image/load_test), which could be useful to you.

deployment/kube/stage/frontend.yaml
```diff
@@ -17,6 +19,15 @@ spec:
         tier: frontend
     spec:
       affinity:
+        podAffinity:
```
scjody: I don't understand what this is doing. Can you please explain or point to some documentation?

deployment/kube/stage/frontend.yaml
```diff
@@ -43,6 +54,13 @@ spec:
         ports:
         - name: http-server
           containerPort: 9091
+        resources:
+         limits:
+           cpu: 600m
+           memory: 1Gi
```
scjody: Why limit to so little memory? The nodes have 3.75 GB available, and only one imageserver pod should be running on each node.

deployment/kube/stage/frontend.yaml
```diff
@@ -43,6 +54,13 @@ spec:
         ports:
         - name: http-server
           containerPort: 9091
+        resources:
+         limits:
+           cpu: 600m
```
scjody: Do we need to limit CPU usage? Why not let the pod use as much CPU as is available?

deployment/kube/stage/hpa.yaml
```diff
@@ -14,5 +14,5 @@ spec:
   # Set this to 3x "min-nodes":
   minReplicas: 3
   # Set this to 3x "max-nodes":
-  maxReplicas: 3
+  maxReplicas: 6
```
scjody: Either the comment needs to be updated, or something else...

deployment/kube/stage/pdb.yaml
```diff
@@ -0,0 +1,9 @@
+apiVersion: policy/v1beta1
+kind: PodDisruptionBudget
+metadata:
+  name: imageserver-pdb
+spec:
+  minAvailable: 2
```
scjody: 👍

deployment/run_server
```diff
@@ -26,7 +26,7 @@ fi
 pkill Xvfb
 pkill node
 
-xvfb-run --auto-servernum --server-args '-screen 0 640x480x24' ./bin/orca.js serve --request-limit=1000 --safe-mode $PLOTLYJS_ARG $@ 1>/proc/1/fd/1 2>/proc/1/fd/2 &
+xvfb-run --auto-servernum --server-args '-screen 0 640x480x24' ./bin/orca.js serve --safe-mode $PLOTLYJS_ARG $@ 1>/dev/stdout 2>/dev/stderr &
```
scjody: Is there a reason to change to `/dev/stdout` and `/dev/stderr`? This wrapper is being run via monit, so `stdout` and `stderr` of the monit process are not necessarily the right place for this output.

scjody:
I'm also concerned by the idea of using preemptible VMs. According to [this document](https://cloud.google.com/kubernetes-engine/docs/how-to/preemptible-vms):

> [...] any or all of your Compute Engine instances might be preempted and become unavailable. There are no guarantees as to when new instances become available.

My understanding of this is that with preemptible instances, we could lose all our nodes and have no replacement nodes. Do you have any sources that contradict this understanding?

mag009:
> I'm also concerned by the idea of using preemptible VMs. According to [this document](https://cloud.google.com/kubernetes-engine/docs/how-to/preemptible-vms):
> 
> > [...] any or all of your Compute Engine instances might be preempted and become unavailable. There are no guarantees as to when new instances become available.
> 
> My understanding of this is that with preemptible instances, we could lose all our nodes and have no replacement nodes. Do you have any sources that contradict this understanding?

Your right about that. Even with Auto-scale there's a chance that we lose all instances at the same time in every zones. Slim change but still. I guess for stage we don't really care if that happen but for prod we can't take that chance. What I'd like to do is to scale with preempt and have 3 min running on none preempt vm's but I guess that should be in a separated issue.

deployment/kube/stage/frontend.yaml
mag009: Actually, I just realised your question was about the podAffinity. 

It's actually wrong it's podAntiAffinity it's to make sure that when it's scale-down it leave the pod on multiple zones.

mag009:
> I still have minor issue, i'm getting connection refused when it add a container so i probably need to adjust the health check.

updated : #41

deployment/kube/stage/frontend.yaml
scjody: OK, that makes more sense, thanks.

deployment/kube/stage/frontend.yaml
scjody: OK, sounds good.

scjody:
Let's stick with regular instances for now. If we run into significant cost issues we can consider preemptible instances, but we don't need it right now. Autoscaling alone should provide significant savings.

Please let me know when you're ready for a re-review on this!

mag009:
@scjody ready for review. 

Just so you know I've created a dedicated pool for the imageserver. The reason is  I want to avoid mixing kube-system with default services. For example, heapster is a critical component for Auto-scale  I ran into issue when I introduced load where heapster stopped responding and the auto-scale stopped.

Procedure to deploy in prod : 
 - create new pool with gcloud named : imageserver 
  - kubectl replace -f kube/prod

deployment/kube/prod/frontend.yaml
```diff
@@ -10,22 +10,28 @@ metadata:
     app: imageserver
 spec:
   replicas: 3
+  strategy:                                                                     
+    rollingUpdate:                                                              
+      maxSurge: 100%                                                            
+      maxUnavailable: 1
```
scjody: Would it be reasonable to make this higher? In streambed we upgrade 25% of our nodes at a time (or that was the intention anyway), and if we lose an availability zone that's 33% of our nodes.

deployment/kube/prod/frontend.yaml
```diff
@@ -10,22 +10,28 @@ metadata:
...
+      maxSurge: 100%                                                            
+      maxUnavailable: 1                                                         
+    type: RollingUpdate  
   template:
     metadata:
       labels:
         app: imageserver
         tier: frontend
     spec:
-      affinity:
```
scjody: Don't we still need antiAffinity to prevent two pods from ending up on the same node? Or are you counting on resource limits to do that? (Smaller commits, and explaining your reasoning in the commit comment would help here...)

deployment/kube/prod/frontend.yaml
```diff
@@ -39,6 +45,12 @@ spec:
         # starting the pod. This is useful when debugging, but should be turned
         # off in production.
         imagePullPolicy: Always
+        resources:                                                              
+         limits:                                                                
+           memory: 2396Mi
```
scjody: Isn't this still a bit small compared to the 3.75 GB available?

deployment/kube/prod/hpa.yaml
```diff
@@ -12,7 +12,7 @@ spec:
     kind: Deployment
     name: imageserver
   # Set this to 3x "min-nodes":
-  minReplicas: 12
-  # Set this to 3x "max-nodes":
+  minReplicas: 3
+  # Set this to 12x "max-nodes":
```
scjody: Can you please explain the reason for this change?

mag009: this is to scale down we want to have a minimum a 3 instances when the load is low.

deployment/kube/prod/frontend.yaml
mag009: 3.75 but 1G is reserved for kubelet and I've tested with 2.5G and I was starting to get some evicted due to not enough ram. We also have fluentd that uses 200M~ so it add up quickly.

2.4G is really the max we can get out of this instance.

deployment/kube/prod/frontend.yaml
mag009: counting on the resource limit to do that.. and if we do switch for larger instances than we won't care if they spin-up to the same server.

deployment/kube/prod/frontend.yaml
mag009: if we have 3 pods running it will spin-up 3 new with the latest image and kill one by one the old one.

mag009:
PR #130 fix the travis-ci

deployment/kube/prod/frontend.yaml
scjody: When I designed this initially I couldn't find a way to set the resource requests such that one and only one imageserver process could occupy a node, but also allow Kubernetes internal pods to occupy that node.

Is there a way to do it now? This will be an issue if we want to have imageservers in the default node pool, and I think we do.

deployment/kube/prod/frontend.yaml
scjody: OK, that's unfortunate but thanks for the explanation.

deployment/kube/prod/hpa.yaml
scjody: I mean changing `maxReplicas` to be 12x "max-nodes".

deployment/kube/prod/frontend.yaml
mag009: yes its either a fixed number or a % so we can probably set it to 50%

deployment/kube/prod/hpa.yaml
mag009: i might have to increase it, we peaked at 16 last night

deployment/kube/prod/frontend.yaml
mag009: its the case now that's why i'm using a new pool like we have for redis. This way kube-system won't be allow to run there so we won't affect our critical pods.

The only way i saw it done was with toleration + taint

I don't see why we would want imageservers to run on the default pool. Any reason?

deployment/kube/prod/frontend.yaml
scjody: 30% would be safer, unless you can guarantee that Kubernetes will wait for all the new nodes to become available before starting to remove nodes. (We wouldn't want to end up with 50% of the required number of nodes.)

deployment/kube/prod/hpa.yaml
scjody: Why are we changing this from 3x "max-nodes" to 12x or 16x "max-nodes"? Unless something else changed, the "max-nodes" variable set in GKE sets the maximum number of nodes per zone, and so with 3 zones we want to multiply this number by 3 to get "maxReplicas".

"minReplicas" works the same way except for "min-nodes". 

If things have changed (as a result of some change elsewhere in GKE, or as a result of your work), please explain what's changed.

deployment/kube/prod/frontend.yaml
scjody: I explained my concerns about adding a new node pool briefly in the last paragraph here: https://github.com/plotly/orca/pull/128#pullrequestreview-156163750

You don't need tolerations and taints to prevent two pods from occupying the same node. That's what the `podAntiAffinity` statement you're removing was doing, and it was working.

mag009: - If I use pod `podAntiAffinity` to prevent imageserver on kube-system. Will end-up with the same having dedicated machine for kube-system but sharing the same pool... I don't mind doing. Just simpler to use a pool "backend" and put everything else related in there.

- If I set an `podAntiAffinity` for imageserver we must make sure to apply it on every app this is where I think a pool make sense.

- Another situation is when it scale down it evict the kube-system service that is on the node and it restart else where in case of heapster we lose 5 min of metrics. Not a big deal but we might end up with lots of gap in our graph

deployment/kube/prod/frontend.yaml
scjody: OK, the default sounds reasonable.

deployment/kube/prod/frontend.yaml
scjody: I'm just suggesting using `podAntiAffinity` to prevent two imageserver pods from occupying the same node like we do now.

I still don't understand why having Kubernetes internal pods on the same nodes as the imageserver pods is a big deal. They don't use significant amounts of resources, do they? I do understand your concern about losing metrics, but I think it's worth trying anyway. I'm surprised Kubernetes isn't designed to scale down by terminating other nodes rather than these, but it sounds like something we have to live with.

deployment/kube/prod/frontend.yaml
```diff
@@ -10,22 +10,50 @@ metadata:
...
+          requiredDuringSchedulingIgnoredDuringExecution:                       
+            nodeSelectorTerms:                                                  
+            - matchExpressions:                                                 
+              - key: failure-domain.beta.kubernetes.io/zone                     
+                operator: In                                                    
+                values:                                                         
+                - us-central1-a                                                 
+                - us-central1-b                                                 
+                - us-central1-c
+        podAffinity:
```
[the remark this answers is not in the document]
mag009: I'm going to check with @etpinard  and make sure the CI pass before merging and deploying.

I've tested with the following examples : https://drive.google.com/open?id=19_bM6OPBQ-T74qZbz32uSpD50DDloXJs and the folder : jody-imageserver-test:/home/scjody/full/

Your right about AntiAffinity.. I just commited the change.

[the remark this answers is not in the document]
etpinard: ref https://github.com/plotly/orca/pull/130#issuecomment-424391137 

#130 should get merged soon, merging master into this branch after that should suffice to get the tests to pass again.

Alternatively, you can cherry-pick the 4 commits off #130 into this branch.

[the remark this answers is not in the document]
scjody: Thanks for the `podAntiAffinity` fix.

etpinard:
Tests are now :white_check_mark: on `master`.

mag009:
@scjody the I've process the success folder and all of them returned a 200, except file that are too large which returned a 400: textPayload":"400 - invalid or malformed request syntax (figure data is likely to make exporter hang, rejecting request.

etpinard:
@mag009 did you test this out using Electron v2 afterall?

mag009:
@etpinard yes, i did but the memory issue is still present in 2.0.9. I've only tested few files not the entire success folder.

etpinard:
Ok great. Well, if the memory issues aren't worse using electron 2.0.9, we should be updating.

@mag009 Can you testing the entire success folder using electron 2.0.9 or write the steps to do so?

PR #130, opened 2018-09-13
Merge electron-builder pack and conda pack jobs

etpinard:
Currently, the conda pack recipe runs `npm i` and `npm run pack` before packing. This has the advantage of making the conda-pack process standalone, but when one runs the electron-build pack and the conda pack jobs in succession (like we do on CI), these two steps are duplicated :turtle: .

This PR makes the conda `recipe/` folder assume that `npm i` and `npm run pack` have been ran before the conda-pack command. This should significantly speed our Travis and AppVeyor CI runs. :racehorse: 

Moreover, for consistency, I merged the conda-pack steps with the electron-builder steps on CircleCI. Here, this makes our config file cleaner, but does add a few seconds of CI time. As the CircleCI workflow is almost always the first to complete, I think simplifying the config here is a win :trophy: 

Oh and this PR (for some reason) also makes Travis pass again. It has been falling all day long with

image

during the `npm run pack` command inside the conda recipe.

-----

@jonmmease does this look ok to you?

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -169,10 +169,10 @@ jobs:
           command: |
             wget http://repo.continuum.io/miniconda/Miniconda3-latest-Linux-x86_64.sh -O miniconda.sh
             chmod +x miniconda.sh
+            ./miniconda.sh -b -p $HOME/miniconda
       - run:
           name: Conda pack
           command: |
-            ./miniconda.sh -b -p $HOME/miniconda
             $HOME/miniconda/bin/conda install --yes conda-build
             $HOME/miniconda/bin/conda build recipe/
             ls $HOME/miniconda/conda-bld/linux-64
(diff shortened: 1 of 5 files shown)


etpinard:
Ping @jonmmease 

I wouldn't want to modify your conda recipes w/o you taking a :eye: at it.

jonmmease:
Sorry I missed this before! Will do by this evening

.circleci/config.yml
```diff
@@ -168,39 +164,27 @@ jobs:
...
-      - run: zip -r artifacts/conda-linux-64.zip ./linux-64
+      - run:
+          name: Install miniconda
+          command: |
+            wget http://repo.continuum.io/miniconda/Miniconda3-latest-Linux-x86_64.sh -O miniconda.sh
+            chmod +x miniconda.sh
+      - run:
+          name: Conda pack
+          command: |
+            ./miniconda.sh -b -p $HOME/miniconda
```
jonmmease: very minor, but this is the command that installs miniconda so it should probably be the last command in the "Install miniconda" block above.

I really like how much cleaner this is with the multi-line `command:` blocks!

etpinard:
Thanks for the review @jonmmease !

PR #135, opened 2018-09-28
add support for CORS

antoinerg:
Adds a `--cors` option to enable Cross-Origin Resource Sharing (CORS).

This may be out of scope for Orca since it can be handled at the HTTP layer by a downstream server but it was also very simple to add here.

### DIFF
--- a/bin/serve.js
+++ b/bin/serve.js
@@ -40,6 +40,10 @@ const OPTS_META = [].concat([{
   name: 'debug',
   type: 'boolean',
   description: 'Starts app in debug mode.'
+}, {
+  name: 'cors',
+  type: 'boolean',
+  description: 'Enables Cross-Origin Resource Sharing (CORS)'
 }])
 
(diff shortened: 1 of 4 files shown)


etpinard:
Very cool. Two comments:

- Would it better to enable CORS by default? Strictly-speaking this would be a breaking change, so the way you have it here is :+1: for v1. I'm interested in hearing your thoughts on this.
- This probably deserves a test. An [integration test](https://github.com/plotly/orca/blob/master/test/integration/orca_serve_test.js) might be nice, but I think an unit similar to [this one](https://github.com/plotly/orca/blob/master/test/unit/server_test.js#L146-L166) would suffice.

antoinerg:
@etpinard I have never seen CORS enabled by default in any HTTP service since it can be a security risk. I think I would leave it off in case we manage sensitive information in Orca down the road.

etpinard:
Beautiful.

:dancer:

PR #144, opened 2018-11-02
K8s resources adjustment

mag009:
part of plotly/streambed#11766

as per below I've miss-understood the CPU metric used by the auto-scale 
It is based of the cpu requests in our case it is very low at 100m and the percentage set is 90% so that mean at 90m we start scaling and by that it start spinning up new node. As per below:

```
kubectl describe hpa

  Normal  SuccessfulRescale  38m (x591 over 36d)   horizontal-pod-autoscaler  New size: 4; reason: cpu resource utilization (percentage of request) above target
  Normal  SuccessfulRescale  14m (x499 over 36d)   horizontal-pod-autoscaler  New size: 5; reason: cpu resource utilization (percentage of request) above target
  Normal  SuccessfulRescale  10m (x98 over 36d)    horizontal-pod-autoscaler  New size: 10; reason: cpu resource utilization (percentage of request) above target
  Normal  SuccessfulRescale  6m22s (x29 over 36d)  horizontal-pod-autoscaler  New size: 14; reason: cpu resource utilization (percentage of request) above target
  Normal  SuccessfulRescale  20s (x1548 over 36d)  horizontal-pod-autoscaler  New size: 3; reason: All metrics below target
```
so in our case we barely use 1400m so by increase the requests to 400m + a CPU target of 150% = 600m * 3  total of 1800m which give us more than what we need. Worst case the auto-scale will kick-in.

### DIFF
--- a/deployment/kube/prod/frontend.yaml
+++ b/deployment/kube/prod/frontend.yaml
@@ -69,7 +69,7 @@ spec:
          limits:                                                                
            memory: 2396Mi                                                       
          requests:                                                              
-           cpu: 100m
+           cpu: 400m
            memory: 1Gi
         ports:
         - name: http-server
--- a/deployment/kube/prod/hpa.yaml
+++ b/deployment/kube/prod/hpa.yaml
@@ -15,4 +15,4 @@ spec:
   minReplicas: 3
   # Set this to 3x "max-nodes":
   maxReplicas: 18
-  targetCPUUtilizationPercentage: 30
+  targetCPUUtilizationPercentage: 150
--- a/deployment/kube/stage/frontend.yaml
+++ b/deployment/kube/stage/frontend.yaml
@@ -11,7 +11,7 @@ metadata:
   labels:
     app: imageserver
 spec:
-  replicas: 3
+  replicas: 1
   strategy:
     rollingUpdate:
       maxSurge: 100%
--- a/deployment/kube/stage/frontend.yaml
+++ b/deployment/kube/stage/frontend.yaml
@@ -74,7 +74,7 @@ spec:
          limits:
            memory: 2396Mi
          requests:
-           cpu: 100m
+           cpu: 400m
            memory: 1Gi
         volumeMounts:
         - mountPath: "/usr/share/fonts/user"
--- a/deployment/kube/stage/hpa.yaml
+++ b/deployment/kube/stage/hpa.yaml
@@ -15,4 +15,4 @@ spec:
   minReplicas: 3
   # Set this to 3x "max-nodes":
   maxReplicas: 6
-  targetCPUUtilizationPercentage: 30
+  targetCPUUtilizationPercentage: 150
--- a/deployment/kube/stage/hpa.yaml
+++ b/deployment/kube/stage/hpa.yaml
@@ -12,7 +12,7 @@ spec:
     kind: Deployment
     name: imageserver
   # Set this to 3x "min-nodes":
-  minReplicas: 3
+  minReplicas: 1
   # Set this to 3x "max-nodes":
-  maxReplicas: 6
+  maxReplicas: 2
   targetCPUUtilizationPercentage: 150


mag009:
@scjody ready for review

scjody:
@mag009 I don't understand your explanation. Can you provide more details on what's happening now and how it needs to change?

It looks like the old target CPU utilization was 30% - where does 90% (in your explanation) come from? I also don't understand why we need to increase the CPU requests from 100m to 400m. Doesn't that number just affect scheduling, in other words what other pods can coexist on the same node as an imageserver pod?

Finally have you done any load testing of the new values to make sure the cluster still scales up when needed?

mag009:
> @mag009 I don't understand your explanation. Can you provide more details on what's happening now and how it needs to change?
> 

From the metric above, we can see how often we scaled. In our case the auto-scale add/remove new node due to the node affinity that prevent from having 2 imageserver of running on the same node.

 - For example : We scaled to 14 replicas 29 times during  last 36 days, which represent  90m / 940m per node (waste of resources).

The goal is to maximized the utilization of our resources and save money

> It looks like the old target CPU utilization was 30% - where does 90% (in your explanation) come from? I also don't understand why we need to increase the CPU requests from 100m to 400m. Doesn't that number just affect scheduling, in other words what other pods can coexist on the same node as an imageserver pod?
> 

I manually changed the value from 30% -> 90% , it was constantly taking up/down nodes and I saw the bill was increasing ( should of been in a pr  ).

- Correct the cpu requests is for scheduling only. We could leave it to 100m and set the TargetCPU to 600% = 60% Utilization per node.

> Finally have you done any load testing of the new values to make sure the cluster still scales up when needed?

I've tested on stage by generating load using `ab -n 10000 -c 20 -p 0.json http://10.128.0.17:9091/`  with single file in parallel and with random sized image to simulate regular traffic.

As we can see the result are much more efficient:  
```
Metrics:                                               ( current / target )
  resource cpu on pods  (as a percentage of request):  457% (457m) / 600%
Min replicas:                                          3
Max replicas:                                          6
Deployment pods:                                       3 current / 3 desired

```
```
kubectl top pods
NAME                           CPU(cores)   MEMORY(bytes)   
imageserver-588b9fcd55-fx4zw   471m         2216Mi          
imageserver-588b9fcd55-rqgj2   420m         1186Mi          
imageserver-588b9fcd55-sv56m   480m         1813Mi  
```
```
kubectl top node
NAME                                   CPU(cores)   CPU%   MEMORY(bytes)   MEMORY%     
gke-stage-default-pool-943c54e3-p8gl   496m         52%    2843Mi          107%        
gke-stage-default-pool-e4669036-6gxh   677m         72%    3064Mi          115%        
gke-stage-default-pool-e4669036-lpdr   393m         41%    2530Mi          95%     
```


this is how prod currently looks like under (load) :

```
Metrics:                                               ( current / target )
  resource cpu on pods  (as a percentage of request):  34% (34m) / 90%
Min replicas:                                          3
Max replicas:                                          18
Deployment pods:                                       10 current / 10 desired
```
```
kubectl top pods
NAME                           CPU(cores)   MEMORY(bytes)   
imageserver-7f65b96d76-229b4   40m          452Mi           
imageserver-7f65b96d76-486tc   26m          594Mi           
imageserver-7f65b96d76-96hsv   27m          430Mi           
imageserver-7f65b96d76-d86ks   102m         1182Mi          
imageserver-7f65b96d76-h7h9x   110m         1807Mi          
imageserver-7f65b96d76-pq5bp   34m          402Mi           
imageserver-7f65b96d76-snbk9   66m          638Mi           
imageserver-7f65b96d76-vc648   92m          1730Mi          
```
```
kubectl top node
NAME                                  CPU(cores)   CPU%   MEMORY(bytes)   MEMORY%   
gke-prod-default-pool-2e5abd50-2b7t   105m         11%    1381Mi          52%       
gke-prod-default-pool-2e5abd50-2s37   80m          8%     1296Mi          49%       
gke-prod-default-pool-2e5abd50-br3w   97m          10%    763Mi           28%       
gke-prod-default-pool-2e5abd50-j9vd   79m          8%     952Mi           35%       
gke-prod-default-pool-e20874d7-6lhh   58m          6%     878Mi           33%       
gke-prod-default-pool-e20874d7-7f4x   65m          6%     875Mi           33%       
gke-prod-default-pool-e20874d7-kgwd   130m         13%    2413Mi          91%       
gke-prod-default-pool-e20874d7-t0w8   67m          7%     773Mi           29%       
gke-prod-default-pool-ed1d55e6-0f5q   184m         19%    2618Mi          99%       
gke-prod-default-pool-ed1d55e6-gs16   158m         16%    1936Mi          73%    
```

scjody:
> I manually changed the value from 30% -> 90% , it was constantly taking up/down nodes and I saw the bill was increasing ( should of been in a pr ).

Yes, this should have been a PR, or at least a discussion with the team. When did you make this change?

> Correct the cpu requests is for scheduling only. We could leave it to 100m and set the TargetCPU to 600% = 60% Utilization per node.

Where does 600% come from? How is that equivalent to 90% in your original explanation? Can you explain exactly what these numbers mean and how you calculated them?

Thanks for the details on your testing. My concern is: with the new settings, are we autoscaling enough to keep up with the load? Do you have any data on this?

mag009:
> > I manually changed the value from 30% -> 90% , it was constantly taking up/down nodes and I saw the bill was increasing ( should of been in a pr ).
> 
> Yes, this should have been a PR, or at least a discussion with the team. When did you make this change?

I made the change on *Sept 29* 2-3 days after releasing it . I saw it was using 18 nodes so i just reacted and increased the %.

> 
> > Correct the cpu requests is for scheduling only. We could leave it to 100m and set the TargetCPU to 600% = 60% Utilization per node.
>
> Where does 600% come from? How is that equivalent to 90% in your original explanation? Can you explain exactly what these numbers mean and how you calculated them?
>

I never said it was equivalent of 90%, although I forgot to mention that we need to increase it in order to maximized the utilization of our nodes. 
  
This is the math I used for that: 
replicas 18 * 90m =  1620m 
replicas 3 * 600m ( 100m with a CPUtarget of 600% = 600m )  = 1800m

> Thanks for the details on your testing. My concern is: with the new settings, are we autoscaling enough to keep up with the load? Do you have any data on this?

This graph represent our highest peak during the last 30 days : 
k8sload

on all 3 pools central-a , central-b and central-c we never reached out more than 20% of cpu utilization.
We can set the CPUtarget to a less aggressive value  : 200%  and re-evaluate in a month ?

scjody:
> I made the change on Sept 29 2-3 days after releasing it .

OK, so it's been at 90% for a while, so we need to increase from that number. Got it.

> This is the math I used for that:
> replicas 18 * 90m = 1620m
> replicas 3 * 600m ( 100m with a CPUtarget of 600% = 600m ) = 1800m

I still don't understand. The calculations alone don't really help me understand the reasoning - _why_ are these the numbers that we want to change to? (Specifically your PR sets CPU requests to `400m` and target CPU utilization to `150%`. I'm OK discussing either these numbers or a different proposal, but I really want to understand where they come from before we make any more changes.)

I'm also still concerned about the safety of these changes. We don't want to end up in a situation where the cluster is at capacity but autoscaling doesn't happen. I thought from the work in #9865 we'd be able to say something like "When the CPU usage of our imageserver pods is over XX%, we need to scale up", and from there we could implement a solution. Would you be able to let us know these numbers? (If you'd still need to put in a lot of  effort to figure out these numbers we could consider increasing the autoscaling parameters gradually and keeping an eye out for problems on prod, but my understanding is you already did the needed tests in #9865 and I'd prefer to take a safer approach based on measurements.)

mag009:
> > I made the change on Sept 29 2-3 days after releasing it .
> 
> OK, so it's been at 90% for a while, so we need to increase from that number. Got it.
> 
> > This is the math I used for that:
> > replicas 18 * 90m = 1620m
> > replicas 3 * 600m ( 100m with a CPUtarget of 600% = 600m ) = 1800m
> 
> I still don't understand. The calculations alone don't really help me understand the reasoning - _why_ are these the numbers that we want to change to? (Specifically your PR sets CPU requests to `400m` and target CPU utilization to `150%`. I'm OK discussing either these numbers or a different proposal, but I really want to understand where they come from before we make any more changes.)
> 

We have a total of 940mCPU per node and a minimum of 3 nodes ( 1 per zone ). Running 24/7
  - A total of 2820mCPU
  - 1391mCPU is allocated to kube-system pods
  - Which leaves us 1429mCPU for the imageserver.

Based on the highest CPU request we had over the last 30 days (~1500mCPU).  As per graph : https://github.com/plotly/orca/pull/144#issuecomment-436082074

This is where I got the 600 * 3 replicas = 1800mCPU. I agree it's a bit too aggressive :smile:  but my goal was to avoid scaling and better utilized the node when we know that this is the only thing running.

> I'm also still concerned about the safety of these changes. We don't want to end up in a situation where the cluster is at capacity but autoscaling doesn't happen. I thought from the work in #9865 we'd be able to say something like "When the CPU usage of our imageserver pods is over XX%, we need to scale up", and from there we could implement a solution. Would you be able to let us know these numbers? (If you'd still need to put in a lot of effort to figure out these numbers we could consider increasing the autoscaling parameters gradually and keeping an eye out for problems on prod, but my understanding is you already did the needed tests in #9865 and I'd prefer to take a safer approach based on measurements.)

 I do have these number as I pointed out above : https://github.com/plotly/orca/pull/144#issue-227995995 these represent the number of time it scaled-up/down. It's hard to measure from this...  It tells me that we scale too often. 

CPU requests for a week : 
avg

We can see that it peaks often explaining the the scale-up/down.

As for #9865 I didn't have any metrics as the auto-scale wasn't setup yet.

mag009:
@scjody what you say if I change the requests CPU to 300mCPU and a target of 90% 
 - it should cover most of our small peaks and avoid us to spin-up new nodes.

Does it sound acceptable to you?

scjody:
@mag009 I agree that we're scaling up too often. Obviously though we need to be concerned with the availability and responsiveness of the service, so we need to make sure we don't go too far in the other direction (in other words not scaling up often enough).

It doesn't sound like we have any measurements or data to support your new proposal though. How much time would it take for you to test autoscaling by doing load testing using stage? Like I said above if it's going to take a significant amount of time then experimenting using prod may be the way to go, but I'd like to consider the tradeoffs first.

mag009:
> @mag009 I agree that we're scaling up too often. Obviously though we need to be concerned with the availability and responsiveness of the service, so we need to make sure we don't go too far in the other direction (in other words not scaling up often enough).
> 
> It doesn't sound like we have any measurements or data to support your new proposal though. How much time would it take for you to test autoscaling by doing load testing using stage? Like I said above if it's going to take a significant amount of time then experimenting using prod may be the way to go, but I'd like to consider the tradeoffs first.

Should take me roughly 2-3h to test.

PR #152, opened 2018-11-27
EMF file support

antoinerg:
This PR is a work in progress adding EMF support to Orca. After testing Inkscape, unoconv and svg2vector I went with Inkscape because it gives the best results.

The biggest challenge with converting SVG to EMF have to do with opacity. Although EMF supports full transparency, [it does not support opacity between 0 and 1](https://stackoverflow.com/questions/22813312/does-emf-enhanced-metafile-support-opacity). This is a problem because:
- We have a lot of rectangles in our figures with `fill: rgba(0,0,0,0)`. Inkscape throws away the alpha channel and is then left with a black rectangle instead of a transparent one! Example:
inkscape emf
- We overlay a **semi-transparent** PNG of WebGL figures on top of SVG axes.
test emf

Finally, Inkscape doesn't handle well the way we define our SVG linear gradients:
inkscape emf

Those issues can be addressed by cleaning the SVG prior to feeding it to Inkscape as I demonstrate here. 

In commit 035ed22, I fix most of the black rectangles by removing transparent rectangles, properly convert the gradient in EMF by changing the `gradientUnits` of our SVG `linearGradient` element and finally manually do alpha compositing for each pixel of the PNG images of WebGL traces.

You can find a comparison of Orca's PNG and Orca's EMF converted to PNG at https://gist.github.com/antoinerg/680ce690b039fbcb49b976de1d28e1c6

To showcase the progress so far, the above figures now looks like this in EMF:
14 emf
gl3d_chrisp-nan-1 emf
basic_heatmap emf


.circleci/config.yml
```diff
@@ -61,7 +61,7 @@ jobs:
           name: Install deps
           command: |
             npm install
-            sudo apt-get install poppler-utils libgconf-2-4
+            sudo apt-get update && sudo apt-get install inkscape poppler-utils libgconf-2-4
```
etpinard: minor :cow2:

```
sudo apt-get update 
sudo apt-get install inkscape poppler-utils libgconf-2-4
```

package-lock.json
```diff
@@ -143,7 +143,7 @@
         "string-width": {
           "version": "2.1.1",
           "resolved": "https://registry.npmjs.org/string-width/-/string-width-2.1.1.tgz",
-          "integrity": "sha512-nOqH59deCq9SRHlxq1Aw85Jnt4w6KvLKqWVik6oA9ZklXLNIOlqg4F2yrT1MVaTjAqvVwdfeZ7w7aCvJD7ugkw==",
+          "integrity": "sha1-q5Pyeo3BPSjKyBXEYhQ6bZASrp4=",
```
etpinard: Hmm. Which npm version are you using?

src/util/inkscape.js
```diff
@@ -0,0 +1,150 @@
...
+      (cb) => childProcess.exec(cmd, cb),
+      (cb) => fs.readFile(outPath, cb)
+    ], (err, result) => {
+      destroyTmpFiles(() => {
+        cb(err, result[2])
+      })
+    })
+  }
+
+  cleanSvg (svg) {
```
etpinard: Oh wow. That's some fancy regex work!

How you tried using something like [`xml2.js`](https://www.npmjs.com/package/xml2js) to loop over all those SVG nodes? It should make things easier to maintain.

etpinard:
@antoinerg Thanks for making https://gist.github.com/antoinerg/680ce690b039fbcb49b976de1d28e1c6 - the results look pretty good!

I've noticed that fill from carpets traces and some contour traces (the ones coloring constraint regions?) appear black. Is that something that can be fixed?

Ideally, I would prefer _not_ using regexes to do all the heavy compatibility work before the SVG -> EMF conversion. Using something like [`xml2js`](https://www.npmjs.com/package/xml2js) to parse the SVG, then looping over all the nodes checking for `fill-opacity`, `stroke-opacity` and `opacity` attributes and then applying the conversion sounds worth the try.

We could even move this compatibility logic to plotly.js' [`Snapshot.toSVG`](https://github.com/plotly/plotly.js/blob/master/src/snapshot/tosvg.js) - which already does a few similar things. I'm not sure if we should be adding a `Plotly.toImage` flag (e.g. `noOpacity`) or add a "format" (e.g. `svg-for-emf`)?

--------

Thinking about this again, I think adding a `noOpacity` flag in plotly.js makes sense. There are probably other image formats out there that don't fully support an alpha channel. Now this doesn't solve all the EMF/Inkscape issues; perhaps these remaining issues could be handling in Orca? What do you think @antoinerg ?

src/util/inkscape.js
antoinerg: I was pondering whether to use regex vs parsing the XML. The latter is certainly less error-prone.

Now that I know what cleaning needs to be done, I will look into `xml2.js`!

antoinerg:
@etpinard thank you for the review!

> I've noticed that fill from carpets traces and some contour traces (the ones coloring constraint regions?) appear black. Is that something that can be fixed?

I decided to wait for your feedback before looking into those. There are also some`gl2d` traces that aren't perfect. I will let you know what we can do with those!

> Ideally, I would prefer not using regexes to do all the heavy compatibility work before the SVG -> EMF conversion. Using something like xml2js to parse the SVG, then looping over all the nodes checking for fill-opacity, stroke-opacity and opacity attributes and then applying the conversion sounds worth the try.

I agree that regexes feel like a hack and are error-prone. I'll look into `xml2js`.

> Thinking about this again, I think adding a noOpacity flag in plotly.js makes sense. There are probably other image formats out there that don't fully support an alpha channel.

I was wondering the same thing: how much cleaning should be done in plotly.js vs Orca. A `noOpacity` flag in plotly.js could be nice and we have a DOM over there: no need for `xml2js`. I need to think about it :thinking: 

> Now this doesn't solve all the EMF/Inkscape issues; perhaps these remaining issues could be handling in Orca?

Yes, some of the fixes for Inkscape's SVG-to-EMF glitches should definitely stay in Orca (ex. changing how we define SVG `linearGradients`) . One good reason for this is that those fixes might depend on Inkscape's version (something we should always actually check in Orca prior to cleaning up!).

package-lock.json
antoinerg: I'm using 6.2.0. I should probably upgrade. What are you using?

etpinard: image

yeah, upgrading would be best.

antoinerg:
> I've noticed that fill from carpets traces and some contour traces (the ones coloring constraint regions?) appear black. Is that something that can be fixed?

### Carpets
After removing the semi-transparent fills, the carpets now look like:
cheater_constraint_greater_than
cheater_constraint_greater_than emf

### Contours

About conversion problems with contours, are you referring to the result below?
contour_nolines emf
contour_nolines

If so, I need to investigate the clipping. I suspect it has to do with the way we define our units (similarly to how we fixed the `linearGradient`.

### gl2d

Unfortunatley, those one are much harder to deal with because they were designed with transparency in mind: they have SVG axes on top of which there are a transparent PNGs of the WebGL contexts. As you know by now, EMF doesn't support transparency very well. Not sure how we could handle this without significant work/modification to the SVG. See this comment for details on EMF and transparency in Inkscape's bug tracker: https://bugs.launchpad.net/inkscape/+bug/1419686/comments/3

antoinerg:
Results from commit 1ecb2f4: https://gist.github.com/antoinerg/e3b7d7b7b97bbff9e0d84c769f0a4e91

antoinerg:
Results for commit 5432b59: https://gist.github.com/antoinerg/1637f08d26664965b6d942e53d56e7a3

etpinard:
@antoinerg Thanks for fixing our tests! 

Now, it looks like the AppVeyor runs have been failing since https://github.com/plotly/orca/pull/152/commits/7dddce534b0996e49c8ee9a06c196fd67ca65cb6 where `jsdom` got bumped down to accomodate Node@6. :rage:

antoinerg:
Commit 212ed21 (which has a bumped down `jsdom`) actually passed tests on AppVeyor so hopefully the problem is not with the new dependencies introduced by this PR. Let's trigger a new build and see if it passes.

antoinerg:
Commit 212ed21 (which has a bumped down `jsdom`) actually passed tests on AppVeyor so hopefully the problem is not with the new dependencies introduced by this PR. Let's trigger a new build and see if it passes.

antoinerg:
Commit 212ed21 (which has a bumped down `jsdom`) actually passed tests on AppVeyor so hopefully the problem is not with the new dependencies introduced by this PR. Let's trigger a new build and see if it passes.

etpinard:
Ok. Something is up with AppVeyor. I pushed an empty commit off master, and AppVeyor fails there too:

https://ci.appveyor.com/project/AppVeyorDashAdmin/image-exporter/builds/20823193

So, I would be ok merging this thing as is. Let me review it one last time.

antoinerg:
@etpinard There's also one last important to do: update the [deployment's Dockerfile](https://github.com/plotly/orca/blob/master/deployment/Dockerfile). Should I submit a separate PR for this one?

etpinard:
> There's also one last important to do: update the [deployment's Dockerfile](https://github.com/plotly/orca/blob/master/deployment/Dockerfile).

Ah good point. Let's do this in this PR!

src/util/inkscape.js
```diff
@@ -0,0 +1,221 @@
...
+    const cmd = `${this.cmdBase} --file ${inPath} --export-emf ${outPath}`
+
+    const destroyTmpFiles = (cb) => parallel([
+      (cb) => fs.unlink(inPath, cb),
+      (cb) => fs.unlink(outPath, cb)
+    ], cb)
+
+    // Get background color from figure's definition
+    var bgColor
+    if (opts.figure.layout.paper_bgcolor) {
```
etpinard: `figure.layout` might not always be defined (unless orca is doing something hacky I don't remember), so I would prefer:

```js
if ((opts.figure.layout || {}).paper_bgcolor) { /* */ }
```

to stay safe.

src/util/inkscape.js
```diff
@@ -0,0 +1,221 @@
...
+      (cb) => fs.readFile(outPath, cb)
+    ], (err, result) => {
+      destroyTmpFiles(() => {
+        cb(err, result[2])
+      })
+    })
+  }
+
+  cleanSvg (svg, bgColor) {
+    const fragment = JSDOM.fragment(svg)
```
etpinard: Nice find! This is probably faster than `DOMParser`

src/util/inkscape.js
```diff
@@ -0,0 +1,221 @@
...
+    })
+
+    // Set fill color to background color if its fill-opacity is 0 but stroke-opacity isn't 0
+    fragment.querySelectorAll('rect').forEach(function (node) {
+      var style = node.getAttribute('style')
+      if (!style) return
+      var m = style.match(fillOpacityZero)
+
+      if (m) {
+        var rgbFill = m[1] + `fill: rgb(${bgColor[0]},${bgColor[1]},${bgColor[2]});`
```
etpinard: :cow2: Let's make this one single ES6 string: `${m[1]} fill: rgb ... ;`

etpinard:
Nice job. The new jsdom-based logic looks fantastic. 

https://gist.github.com/antoinerg/1637f08d26664965b6d942e53d56e7a3 is giving me 500 though, is that just Github acting up?

antoinerg:
> https://gist.github.com/antoinerg/1637f08d26664965b6d942e53d56e7a3 is giving me 500 though, is that just Github acting up?

It's probably Github acting up because the gist is large. Maybe git clone it:
``` 
git clone git@gist.github.com:1637f08d26664965b6d942e53d56e7a3.git
```

By the way, I really like `sxiv` to check the images. `sxiv *.png` and the nkeys `n` and `p` to see the next and previous image. Press `enter` to see preview of many images.

src/util/inkscape.js
antoinerg: Fixed in e76c52b

deployment/Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM ubuntu:xenial
+FROM ubuntu:bionic
```
etpinard: Hmm. We'll have to test that thoroughly on stage. Is this required?

antoinerg: It's definitely the path of least resistance to get the latest version of Inkscape (0.92.3 which is also the version I used in my testing).

deployment/Dockerfile
```diff
@@ -90,7 +90,7 @@ RUN apt-get update -y && apt-get install -y subversion && \
 RUN wget -q -O - https://dl-ssl.google.com/linux/linux_signing_key.pub | apt-key add - && \
     sh -c 'echo "deb [arch=amd64] http://dl.google.com/linux/chrome/deb/ stable main" >> /etc/apt/sources.list.d/google-chrome.list' && \
     apt-get update -y && \
-    apt-get install -y google-chrome-stable xvfb poppler-utils git && \
+    apt-get install -y google-chrome-stable xvfb poppler-utils inkscape git && \
```
etpinard: Have you tried running that locally?

If I remember correctly:

```
docker build . -f deployment/Dockerfile -t tmp-inkscape

docker run tmp-inkscape -d -p 9010:9010 tmp-inkscape

# then
curl localhost:9010 ...
```

antoinerg: Default port is 9091

deployment/Dockerfile
etpinard: Using a ppa shouldn't be too bad:

http://ubuntuhandbook.org/index.php/2017/01/install-inkscape-0-92-ppa-ubuntu-16-04-16-10-14-04/

----

I'm not against bumping Ubuntu, but we can't just casually bump it here w/o some thorough manual testing.

deployment/Dockerfile
antoinerg: The following works for me!
```
docker build . -f deployment/Dockerfile -t tmp-inkscape

# check if ok with
docker images

docker run tmp-inkscape -d -p 9091:9091 tmp-inkscape

# then
cd plotly/plotly.js/test/image/mocks/
curl _XPOST -d @14.json localhost:9091 > /tmp/14.png
```

deployment/Dockerfile
antoinerg: I understand your concern. Let me try to compile Inkscape or install from a deb package.

antoinerg: > Using a ppa shouldn't be too bad:

That did the trick thank you :+1:  

Commit b6daae1 reverted our container to `ubuntu:xenial` and install Inkscape from a PPA

src/util/inkscape.js
antoinerg: Change made in commit e4ab53d

etpinard:
> Results in PNG for latest commit [e76c52b](https://github.com/plotly/orca/commit/e76c52be7e219dcb41737c26f96fd0efc2e3dccd):
> 
> ```
> git clone git@gist.github.com:632c26b518988f63bb766453bf850c9c.git
> ```

Thanks!

The results are very impressive!

Apart from all gl2d and splom mocks, I noticed:

- we could probably improve the look of rangesliders e.g. we could try removing the shaded layers completely instead of having them show up in black
- `category_autorange` has shapes with `fill-color: rgb(0,0,0)` and `fill-opacity: 0`, sounds like we could simply remove `fill-color` here to get better results
- `contour_log` and `contour_autorange` are off like you mentioned in https://github.com/plotly/orca/pull/152#issuecomment-443831680
- `legend_horizontal_autowrap` has a border around the legend items
- all sankey mocks look odd

All in all, out of 600+ mocks, that's not bad.

I'm ready to :dancer:

@antoinerg you can either:

- try to fixup a few of the items above before merging, or
-  merge this and open a new orca issue pasting that list of open items.

---

About the AppVeyor failure, we should try running the `conda pack` steps on a Windows laptop.

antoinerg:
> @antoinerg you can either:
> 
> * try to fixup a few of the items above before merging, or
> * merge this and open a new orca issue pasting that list of open items.

@etpinard, I think I will go with the second course of action so others can benefit as soon as possible from the tests that we fixed in this branch. For example this PR @tarzzz just submitted https://github.com/plotly/orca/pull/165 fails :(

> About the AppVeyor failure, we should try running the conda pack steps on a Windows laptop.

Is this blocking? If so, I will test it first thing tomorrow!

Thank you for the thorough review and for finding all the discrepancies :male_detective:! I will fix them in a separate issue as you suggested :tada:

etpinard:
> Is this blocking? I

Not blocking for this PR, but we should try to fix it ASAP to not delay any upcoming orca releases.

Issue #131, opened 2018-09-16
xvfb-run: error: Xvfb failed to start


lobachevzky:
I am able to reproduce the error with the following minimal code:
```
import plotly.io as pio
import numpy as np
import plotly.graph_objs as go


N = 100
x = np.random.rand(N)
y = np.random.rand(N)
colors = np.random.rand(N)
sz = np.random.rand(N)*30

fig = go.Figure()
fig.add_scatter(x=x,
                y=y,
                mode='markers',
                marker={'size': sz,
                        'color': colors,
                        'opacity': 0.6,
                        'colorscale': 'Viridis'})

pio.write_image(fig, 'plot.png')
```
To which I get the following error:
```
❯ python orca_bug.py 
Traceback (most recent call last):
  File "orca_bug.py", line 21, in <module>
    pio.write_image(fig, 'plot.png')
  File "/home/ethanbro/virtualenvs/598/lib/python3.6/site-packages/plotly/io/_orca.py", line 1492, in write_image
    validate=validate)
  File "/home/ethanbro/virtualenvs/598/lib/python3.6/site-packages/plotly/io/_orca.py", line 1263, in to_image
    ensure_server()
  File "/home/ethanbro/virtualenvs/598/lib/python3.6/site-packages/plotly/io/_orca.py", line 1134, in ensure_server
    validate_executable()
  File "/home/ethanbro/virtualenvs/598/lib/python3.6/site-packages/plotly/io/_orca.py", line 1001, in validate_executable
    returncode=p.returncode))
ValueError: 
The orca executable is required in order to export figures as static images,
but the executable that was found at '/usr/local/bin/orca'
does not seem to be a valid plotly orca executable. Please refer to the end of
this message for details on what went wrong.

If you haven't installed orca yet, you can do so using conda as follows:

    $ conda install -c plotly plotly-orca

Alternatively, see other installation methods in the orca project README at
https://github.com/plotly/orca.

After installation is complete, no further configuration should be needed. 

If you have installed orca, then for some reason plotly.py was unable to
locate it. In this case, set the `plotly.io.orca.config.executable`
property to the full path of your orca executable. For example:

    >>> plotly.io.orca.config.executable = '/path/to/orca'

After updating this executable property, try the export operation again.
If it is successful then you may want to save this configuration so that it
will be applied automatically in future sessions. You can do this as follows:

    >>> plotly.io.orca.config.save() 

If you're still having trouble, feel free to ask for help on the forums at
[...]

PR #153, opened 2018-11-28
Add -a flag to prevent startup error with xvfb

Ejdamm:
The problem is explained in #131


antoinerg:
Thanks @Ejdamm! I agree that suggesting using `-a` flag will probably lead to a better user-experience. :dancer:

PR #156, opened 2018-11-28
Temporarily scale to 51 pods

scjody:
Part of https://github.com/plotly/streambed/issues/11932

See https://github.com/plotly/streambed/issues/11932#issuecomment-442579644 for rationale. I went with 51 to allow a consistent number of nodes per zone.

I also plan on modifying the existing configuration for `default-pool` in the `prod` cluster (PlotlyCloud project):

screen shot 2018-11-28 at 15 07 08
screen shot 2018-11-28 at 15 07 35

@mag009 Please review
@plotly/devops FYI

PR #163, opened 2018-12-03
Revert "Temporarily scale to 51 pods"

scjody:
Reverts plotly/orca#156

@nicolaskruchten and I discussed this on Friday. I'd like to revert this PR and set the node pool autoscaling parameters back the way they were (autoscaling on, max 6 nodes per zone) in the web UI.

Rationale:
* This change wasn't sufficient to prevent downtime (the user reached their throttle and then started hitting our backend sufficiently hard to crash it) and is expensive to keep in place.
* The user has agreed not to hit our site anymore.

@plotly/devops Any objections?
@mag009 Please review.

### DIFF
--- a/deployment/kube/prod/hpa.yaml
+++ b/deployment/kube/prod/hpa.yaml
@@ -11,8 +11,8 @@ spec:
     apiVersion: extensions/v1beta1
     kind: Deployment
     name: imageserver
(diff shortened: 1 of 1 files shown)


mag009:
:dancer:

PR #168, opened 2018-12-10
cache custom fonts

BRONSOLO:
addresses remaining orca work for: https://github.com/plotly/streambed/issues/11328

 - [x] on-prem (tested all image formats ✔️ )

@scjody please review. 

**Note:**  Ultimately, all that was needed for Orca to be made aware of the new fonts was for the fonts to be added to the system font cache via `fc-cache`. 

**Important:** When testing, I discovered that the code being added is not strictly necessary because all fonts within `/usr/share/fonts` will be automatically added to the system font cache when the container starts up. I investigated this and I suspect what is happening is that the system font configuration file `/etc/fonts/fonts.conf` sets the font directories to be: 

```
<dir>/usr/share/fonts</dir>
<dir>/usr/local/share/fonts</dir>
```
and any application that uses the `fontconfig` library will update the font cache as needed (so `fc-cache` will be executed on `/usr/share/fonts` at that time). @etpinard any additional insight here? 

In any case, I think it's worth explicitly calling it ourselves as is done in the code change of this PR.

### DIFF
--- a/deployment/run_server
+++ b/deployment/run_server
@@ -2,6 +2,10 @@
 
 fc-cache -v /usr/share/fonts/user
 
+if [[ $PLOTLY_CUSTOM_FONTS_ENABLED = true ]]; then
+  fc-cache -v /usr/share/fonts/customer
+fi
+
 BUILD_DIR=/var/www/image-exporter/build
 if [[ -n "${PLOTLY_JS_SRC}" ]]; then
   # Fetch plotly js bundle and save it locally:


BRONSOLO:
PDF test: [font_test.pdf](https://github.com/plotly/orca/files/2664512/font_test.pdf)

nicolaskruchten:
Nice test :) Did you make that through the `/create` UI?

nicolaskruchten:
Ah never mind, I see this is the `orca` repo :)

BRONSOLO:
aha Thanks @nicolaskruchten. Yes it was actually. I'm going to make a note in the RCE PR that everything looks 👌 after initial testing.

BRONSOLO:
@nicolaskruchten and @scjody one other thing that came up here is that the font name that is sent to orca via the figure JSON must correspond to the name recognized by the system (the name listed when running `fc-list` within the imageserver container). As such, the font name that is included in `custom_fonts.json` and `custom_fonts.css` must also match this name. In my experience, these names are typically the names associated with the **.tff** file (Comic Sans MS, in my example). I don't see this being an issue as long as it is well documented. I added "documenting the custom font upload flow" as an item here https://github.com/plotly/streambed/issues/11328#issuecomment-413208584. Any thoughts/concerns?

PR #170, opened 2018-12-10
Add env var for request limit

gzork:
Part of https://github.com/plotly/streambed/issues/11979

Adding an ENV VAR for request limit as we need it but only for On-Prem

Tested on:
 * [x] Local On-Prem Instance

@mag009 a quick review please ?


mag009:
:dancer:

PR #176, opened 2018-12-28
Fix --output `orca graph` CLI option for path/to/filename

### LINKED ISSUE #101: orca() does not support file paths
I'm trying to export a static image in R using orca(), but it ignores the file path and just exports it to the current location of the script that I'm running. 

For example, if I run `orca(image, "/images/plot1.png")` from a script located in  `/code/script.R`, it saves it to `/code/plot1.png` instead of where I specified.

etpinard:
fixes #101 and adds a bunch more `orca --graph` tests

cc @antoinerg

### DIFF
--- a/bin/graph.js
+++ b/bin/graph.js
@@ -101,8 +101,12 @@ function main (args) {
     process.exit(0)
   }
 
-  if (!fs.existsSync(opts.outputDir)) {
-    fs.mkdirSync(opts.outputDir)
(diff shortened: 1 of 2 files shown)


etpinard:
Argg. Something is wrong again with `docker-build-and-push`:

image

antoinerg:
Thank you @etpinard for the fix! :dancer:

PR #177, opened 2018-12-29
Bump deps and dev-deps 2018-12-28

etpinard:
Bumping deps (other than `electron` and `electron-builder` - which will require a PR of their own) that still support Node v6 and our dev deps.

cc @antoinerg


antoinerg:
Looks good :dancer:

Issue #173, opened 2018-12-19
Document imageserver image name for maintenance release


BRONSOLO:
All imageserver work for On-Prem 2.5.0 lands on https://github.com/plotly/orca/tree/v1.0.0-onprem. In order to make a maintenance release, we need some way to have a unique name for the updated image. 

**Suggestions:** 

1) Create a new branch for maintenance release v1.0.1-onprem
2) Tag image in quay. 

Option 2 seems to be a nice solution imo. 

I think we should document this for future releases. 

@plotly/devops

scjody:
Managing ORCA versions is up to the folks that work on ORCA: @etpinard and now @antoinerg. For 2.5.0 @etpinard created the `v1.0.0-onprem` tag, and I'm not sure what his plans were for maintenance releases.

Maintenance releases need to be based on the previous version (`v1.0.0-onprem` in this case) and need to contain _only_ the needed fix(es). So normally they'd be built from a maintenance branch (the devops team uses branches of the form `2.5-release`).

scjody:
@BRONSOLO and I investigated this together. It turns out that `v1.0.0-onprem` is a branch in GitHub, not a tag. When https://github.com/plotly/orca/pull/151 was merged to the `v1.0.0-onprem` Orca branch, CircleCI built the a Docker image, pushed it to Quay, and [tagged it as `v1.0.0-onprem`](https://github.com/plotly/orca/blob/eda04a4b0e6b0eba03d38b3ece79f281fb625fab/.circleci/config.yml#L127-L133).

As of this point, any customers who performed a new installation of 2.5.x or updated to any 2.5.x release ended up using the new Orca release rather than the one we were previously shipping. We don't want to do this - Orca releases should be tested as part of an overall On-Prem release before being sent to customers - but by the time we realized the error @BRONSOLO _had_ tested the new On-Prem maintenance release (including the new Orca release). We therefore decided to leave the `v1.0.0-onprem` tag as-is in Quay until we could discuss things in this issue.

(We have created a `v1.0.1-onprem` tag pointing to the new Orca, which @BRONSOLO will use as part of the 2.5.3 maintenance release he's putting out right now.)

Here's what things looked like before yesterday:

![screen shot 2018-12-19 at 16 18 05](https://user-images.githubusercontent.com/1136329/50248632-aee35080-03a9-11e9-83af-fbd33b0198c3.png)

Here's what things looked like yesterday:

[...]

Issue #106, opened 2018-07-30
Use special docker tag on `git push --tags`


etpinard:
See https://circleci.com/docs/2.0/workflows/#git-tag-job-execution

This should make our https://quay.io/repository/plotly/image-exporter?tab=tags a little cleaner.

etpinard:
Done in #179 

![image](https://user-images.githubusercontent.com/6675409/51871502-7fb6ec00-2324-11e9-8372-3171358b0c9b.png)

`v1.2.0` is looking good on quay.io.

PR #179, opened 2018-12-29
Use `git log -1` to determine if version commit during docker-build-and-push job

etpinard:
2nd attempt at https://github.com/plotly/orca/issues/106 after the failed https://github.com/plotly/orca/pull/118

**Note:** I haven't tried to `git push --tags` (is there a test that w/o actually adding a git tags?) 

See:
- https://discuss.circleci.com/t/git-tag-deploys-in-2-0/9493/4 

for more info

cc @antoinerg and https://github.com/plotly/orca/issues/173

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -96,7 +96,7 @@ jobs:
       - image: circleci/node:8.11-stretch-browsers
     steps:
       - setup_remote_docker:
-          reusable: true
+          docker_layer_caching: true
       - restore_cache:
           key: v1-source-{{ .Branch }}-{{ .Revision }}
           keys:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -129,9 +129,15 @@ jobs:
             # tag and push as branch name
             docker tag quay.io/plotly/image-exporter:$CIRCLE_SHA1 quay.io/plotly/image-exporter:$CIRCLE_BRANCH
             docker push quay.io/plotly/image-exporter:$CIRCLE_BRANCH
-            if [[ ! -z "${CIRCLE_TAG}" ]]; then
-              docker tag quay.io/plotly/image-exporter:$CIRCLE_SHA1 quay.io/plotly/image-exporter:$CIRCLE_TAG
-              docker push quay.io/plotly/image-exporter:$CIRCLE_TAG
+
+            # tag and push as version number
+            # inspired by: https://discuss.circleci.com/t/git-tag-deploys-in-2-0/9493/4
+            if git log -1 --pretty=%B | grep "v^[0-9]\+\.[0-9]\+\.[0-9]\+$"; then
+              version=$(git log -1 --pretty=%B)
+              docker tag quay.io/plotly/image-exporter:$CIRCLE_SHA1 quay.io/plotly/image-exporter:$version
+              docker push quay.io/plotly/image-exporter:$version
+            else
+              echo "Not a release, don't push version-tagged image to quay.io"
             fi
       - save_cache:
           paths:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -18,13 +18,13 @@ workflows:
             - test-node-v6
             - test-node-v8
           filters:
-              branches:
-                  # Do not run on branches based off forks as
-                  # this job requires plotly's quay.io credentials.
-                  # Note that branches based off forks are labels pull/???/ on CI.
-                  # Solution taken from:
-                  # https://discuss.circleci.com/t/create-separate-steps-jobs-for-pr-forks-versus-branches/13419/4
-                  only: /^(?!pull\/).*$/
+            branches:
+              # Do not run on branches based off forks as
+              # this job requires plotly's quay.io credentials.
+              # Note that branches based off forks are labels pull/???/ on CI.
+              # Solution taken from:
+              # https://discuss.circleci.com/t/create-separate-steps-jobs-for-pr-forks-versus-branches/13419/4
+              only: /^(?!pull\/).*$/
       - electron-pack-and-release:
           requires:
             - test-node-v6
@@ -97,7 +97,6 @@ jobs:
     steps:
       - setup_remote_docker:
           reusable: true
-
       - restore_cache:
           key: v1-source-{{ .Branch }}-{{ .Revision }}
           keys:
@@ -125,8 +124,10 @@ jobs:
           name: Push Docker image to Quay
           command: |
             docker login -u $DOCKER_USER -p $DOCKER_PASS quay.io
-            docker tag quay.io/plotly/image-exporter:$CIRCLE_SHA1 quay.io/plotly/image-exporter:$CIRCLE_BRANCH
             docker push quay.io/plotly/image-exporter:$CIRCLE_SHA1
+
+            # tag and push as branch name
+            docker tag quay.io/plotly/image-exporter:$CIRCLE_SHA1 quay.io/plotly/image-exporter:$CIRCLE_BRANCH
             docker push quay.io/plotly/image-exporter:$CIRCLE_BRANCH
             if [[ ! -z "${CIRCLE_TAG}" ]]; then
               docker tag quay.io/plotly/image-exporter:$CIRCLE_SHA1 quay.io/plotly/image-exporter:$CIRCLE_TAG


antoinerg:
> Note: I haven't tried to git push --tags (is there a way to test that w/o actually adding a git tags?)

Probably not but I may be wrong. We should make sure this works before merging. Can you simply tag e4d6936 to `v0.0.0` and make sure it gets pushed to Quay?

.circleci/config.yml
```diff
@@ -125,12 +124,20 @@ jobs:
...
+            # tag and push as branch name
+            docker tag quay.io/plotly/image-exporter:$CIRCLE_SHA1 quay.io/plotly/image-exporter:$CIRCLE_BRANCH
             docker push quay.io/plotly/image-exporter:$CIRCLE_BRANCH
-            if [[ ! -z "${CIRCLE_TAG}" ]]; then
-              docker tag quay.io/plotly/image-exporter:$CIRCLE_SHA1 quay.io/plotly/image-exporter:$CIRCLE_TAG
-              docker push quay.io/plotly/image-exporter:$CIRCLE_TAG
+
+            # tag and push as version number
+            # inspired by: https://discuss.circleci.com/t/git-tag-deploys-in-2-0/9493/4
+            if git log -1 --pretty=%B | grep "v^[0-9]\+\.[0-9]\+\.[0-9]\+$"; then
```
antoinerg: The problem discussed in #173 is that for deployment we used Docker tags named after branches instead of tags. We need a clear way of distinguishing the two.

It seems like @scjody and you use the convention that branches do not have a PATCH version in their names but tags do. Is that really the case?

.circleci/config.yml
```diff
@@ -125,12 +124,20 @@ jobs:
           name: Push Docker image to Quay
           command: |
             docker login -u $DOCKER_USER -p $DOCKER_PASS quay.io
-            docker tag quay.io/plotly/image-exporter:$CIRCLE_SHA1 quay.io/plotly/image-exporter:$CIRCLE_BRANCH
             docker push quay.io/plotly/image-exporter:$CIRCLE_SHA1
+
+            # tag and push as branch name
+            docker tag quay.io/plotly/image-exporter:$CIRCLE_SHA1 quay.io/plotly/image-exporter:$CIRCLE_BRANCH
```
antoinerg: If `$CIRCLE_BRANCH` has the same name as an existing git tag, it will overwrite the existing image. This seems error-prone to me. Maybe we could tag the image `quay.io/plotly/image-exporter:branch-$CIRCLE_BRANCH`? See my next comment for the rationale behind this.

etpinard: Good point. Thanks!

> Maybe we could tag the image quay.io/plotly/image-exporter:branch-$CIRCLE_BRANCH? 

I'm a fan of this. Maybe on-prem devs have a strong opinion against it?

etpinard:
Thanks very much for the review @antoinerg !

> We should make sure this works before merging. Can you simply tag [e4d6936](https://github.com/plotly/orca/commit/e4d69362788e7732cfcb7cca707f3ec05e73d79f) to `v0.0.0` and make sure it gets pushed to Quay?

Not sure what's best:

- merge this, release `v1.2.0` (make a `v1.2.0` git tag in the process) and see if this work, or
- make a funny tag `v0.0.0` (that will leave in git history forever) and see of this work.

I prefer the first option, but maybe I should look up how to make `git tag -a` dry runs.

.circleci/config.yml
scjody: I don't understand what you mean by "a PATCH version".

The convention used by the code managed by the devops team is:
* Release branches are of the form `3.0-release`, where `3.0` is the major.minor version of the release. (We don't follow semver but that's not relevant here - the convention works even if semver is used.)
* We add annotated tags to this branch to represent anything we release to customers.
* We add tags to Docker images using Quay's interface as part of adding them to an On-Prem release.

Feel free to adopt any of this that's useful to you and @etpinard, or come up with your own conventions provided they meet the following requirements:
* You're able to give us an Orca release using a Docker image tag that points to an unchanging image.
* Any Orca release that forms part of an On-Prem release can be minimally updated as part of a maintenance release. In other words we can address minor issues without changing any other parts of the code or any underlying dependencies.

scjody:
General comments:
* I still think adding `branch-` to Docker tags that represent branches is strange and unnecessary, but I don't object to you adding that if you feel it's better.
* Testing this with a "funny tag" is definitely useful. I suggest adding some text to the tag to remind people that it's temporary (e.g. `v1.2.0-taggingtest`), and deleting the tag from GitHub when you're done with it if possible (I'm pretty sure it is).

etpinard:
**Update:**

I messed up a few things in my comment:

> **Note:** I haven't tried to `git push --tags` (is there a way to test that w/o actually adding a git tags?)

was completely off. We don't have to `git push --tags` to test the new logic. That new `git log -1` piece isn't activated during `git push --tags`, it is activated during version commits (e.g. following `npm version`). So testing that in

https://github.com/plotly/orca/compare/docker-build-and-push-with-tag-test

(after https://github.com/plotly/orca/pull/179/commits/011cad64441fdccc40b84183e3a910c52a9fe6ca) worked:

image

image

Note that according to our [contributing guidelines](https://github.com/plotly/orca/blob/master/CONTRIBUTING.md#checklist), version commits are made via `npm version` on the `master` branch.

---

This doesn't solve the issue of potential collisions between existing `[0-9].[0-9].[0-9]` tagged images and images from branches named `[0-9].[0-9].[0-9]`. Is appears unlikely that a dev would name a branch e.g, `1.2.0` (`1.2.0-onprem` would be ok), but since this could mean replacing an existing tagged image, I'm thinking of skipping:

```
docker tag quay.io/plotly/image-exporter:$CIRCLE_SHA1 quay.io/plotly/image-exporter:$CIRCLE_BRANCH
```

when `$CIRCLE_BRANCH` passes regex `"^[0-9]\+\.[0-9]\+\.[0-9]\+$"`. @antoinerg thoughts?

antoinerg:
Skipping the tag/push operation when `$CIRCLE_BRANCH` looks like a version number is brilliant:
1) it aligns with @scjody's naming scheme
2) it prevents overwriting existing releases

etpinard:
> Skipping the tag/push operation when `$CIRCLE_BRANCH` looks like a version number

done in -> https://github.com/plotly/orca/pull/179/commits/ed99da06d65c070e5057f7ccc0cd7a9e89abdb48

.circleci/config.yml
```diff
@@ -125,12 +124,29 @@ jobs:
...
-            docker tag quay.io/plotly/image-exporter:$CIRCLE_SHA1 quay.io/plotly/image-exporter:$CIRCLE_BRANCH
             docker push quay.io/plotly/image-exporter:$CIRCLE_SHA1
-            docker push quay.io/plotly/image-exporter:$CIRCLE_BRANCH
-            if [[ ! -z "${CIRCLE_TAG}" ]]; then
-              docker tag quay.io/plotly/image-exporter:$CIRCLE_SHA1 quay.io/plotly/image-exporter:$CIRCLE_TAG
-              docker push quay.io/plotly/image-exporter:$CIRCLE_TAG
+
+            # Is this a "version" commit? e.g. from `npm version`
+            # inspired by: https://discuss.circleci.com/t/git-tag-deploys-in-2-0/9493/4
+            if git log -1 --pretty=%B | grep "^[0-9]\+\.[0-9]\+\.[0-9]\+$"; then
```
antoinerg: This regex wouldn't match anything starting with a `v`. I think it should, shouldn't it?

antoinerg: This regex wouldn't match anything starting with a v. I think it should, shouldn't it?

etpinard: ... but `npm version` writes version commits w/o a leading `v`

etpinard: Right, git tags start with `v`, but version commits like this one:

https://github.com/plotly/orca/commit/d50563f4d6dbc20c4a5a37e15c9465f90e97a37a

don't

antoinerg: Oh wow, I was really confused. `git log --pretty=%B` actually prints out the commit message NOT the git tag. This means a commit message that looks like a version number can screw things up but I guess that won't happen.

Non-blocking but the [release checklist](https://github.com/plotly/orca/blob/master/CONTRIBUTING.md#checklist) states that running `npm version` does a `git commit` with message `vX.Y.Z`. Actually the message for version commits doesn't have a leading `v` and has a message `X.Y.Z`. Maybe fix that typo in `CONTRIBUTING.md` for clarity.

etpinard: > Maybe fix that typo in `CONTRIBUTING.md` for clarity.

Good call. Thanks! Done in -> https://github.com/plotly/orca/commit/0afcec0e15be168f77f91b60750a03a7476ffa89

antoinerg:
Looks good to me! :dancer:

Issue #187, opened 2019-01-11
Add Docker in installation methods


antoinerg:
I think we should advertise our Docker images in the installation methods as Docker is getting more and more popular and *should* solve all the problems related to missing dependencies or misconfiguration.

etpinard:
Good idea! We could either upload a docker image for each of our releases alongside the mac/windows/linux executables:

![image](https://user-images.githubusercontent.com/6675409/51118711-0923d580-17df-11e9-9b89-423784383e96.png)

or just provide a link to quay,io (are those public links?) in the README.

cc https://github.com/plotly/orca/issues/178

austinbutler:
Having it available in the Docker Hub would be great. I saw the somewhat convoluted possible install options in the README and at first thought those were the only options. Then I noticed there's a Dockerfile in the deployment folder in this repo, but even then since it's not documented figured it wouldn't be good to rely on it.

etpinard:
Good point by @antoinerg in https://github.com/plotly/orca/pull/190

> For #187, we may want to write a new Dockerfile aimed at offline use without streambed plotly.js but the public bundle.

where I'd say we _should_ add another Dockerfile!

antoinerg:
> Having it available in the Docker Hub would be great.

@austinbutler Is there something special about Docker Hub or would any other Docker registries fit your needs? We were thinking of using quay.io instead. You can pull from it the same way you pull from Docker Hub.

austinbutler:
Quay would be fine, too. But in terms of visibility Docker Hub is just where I default to when looking for images.

etpinard:
> Can we start it from the most recent LTS version of Ubuntu

yes please.

> I could reuse the ones I made many when I worked on delivering

sure.

PedroRegisPOAR:
I would love to see Docker images to run this!

I have a question, @antoinerg I am not sure if I am in the right place, but as an user I would love to use full `plotly.py` functionalities, that includes `orca`, in an Docker image. Who may provide a way to do this? The [plotly.py](https://github.com/plotly/plotly.py) or the `orca` repository?

I guess that for most users, like me, would expect to have a way do install all `plotly.py` stuff, that includes `orca`, from a Python official base image. Is here the right place to talk about this?

I think if `orca` provide a way to build an image it is going to help to build a image for all `plotly.py` stuff.

[...]

PR #190, opened 2019-01-15
download latest plotly.js library into Docker for offline use

antoinerg:


antoinerg:
@etpinard I didn't remember that. I'll close this one.

For #187, we may want to write a new Dockerfile aimed at offline use without streambed plotly.js but the public bundle.

PR #191, opened 2019-01-15
pass command line options to plotly-dash-preview

antoinerg:
Prior to this PR, the HTML file used by `plotly-dash-preview` would ALWAYS link to plotly.js on the CDN even when the command-line arguments `--plotlyjs` is used.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -39,7 +39,7 @@
     "@sinonjs/commons": {
       "version": "1.3.0",
       "resolved": "https://registry.npmjs.org/@sinonjs/commons/-/commons-1.3.0.tgz",
-      "integrity": "sha1-UKJ1QBa28wqZTO2m2aCow2rdqEk=",
+      "integrity": "sha512-j4ZwhaHmwsCb4DlDOIWnI5YyKDNMoNThsmwEpfHx6a1EpsGZ9qYLxP++LMlmBRjtGptGHFsGItJ768snllFWpA==",
       "dev": true,
       "requires": {
         "type-detect": "4.0.8"
@@ -48,7 +48,7 @@
     "@sinonjs/formatio": {
       "version": "3.1.0",
       "resolved": "https://registry.npmjs.org/@sinonjs/formatio/-/formatio-3.1.0.tgz",
-      "integrity": "sha1-asnR6xghmE2ExJlnJuRdFkbYzOU=",
+      "integrity": "sha512-ZAR2bPHOl4Xg6eklUGpsdiIJ4+J1SNag1DHHrG/73Uz/nVwXqjgUtRPLoS+aVyieN9cSbc0E4LsU984tWcDyNg==",
       "dev": true,
       "requires": {
         "@sinonjs/samsam": "^2 || ^3"
@@ -57,7 +57,7 @@
     "@sinonjs/samsam": {
       "version": "3.0.2",
       "resolved": "https://registry.npmjs.org/@sinonjs/samsam/-/samsam-3.0.2.tgz",
-      "integrity": "sha1-ME+zO9VYWgst+KTIAfy0f6hNjkM=",
+      "integrity": "sha512-m08g4CS3J6lwRQk1pj1EO+KEVWbrbXsmi9Pw0ySmrIbcVxVaedoFgLvFsV8wHLwh01EpROVz3KvVcD1Jmks9FQ==",
       "dev": true,
       "requires": {
         "@sinonjs/commons": "^1.0.2",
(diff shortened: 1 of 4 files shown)


package-lock.json
```diff
@@ -57,7 +57,7 @@
     "@sinonjs/samsam": {
       "version": "3.0.2",
       "resolved": "https://registry.npmjs.org/@sinonjs/samsam/-/samsam-3.0.2.tgz",
-      "integrity": "sha512-m08g4CS3J6lwRQk1pj1EO+KEVWbrbXsmi9Pw0ySmrIbcVxVaedoFgLvFsV8wHLwh01EpROVz3KvVcD1Jmks9FQ==",
+      "integrity": "sha1-ME+zO9VYWgst+KTIAfy0f6hNjkM=",
```
etpinard: This stuff here shouldn't happen.

Which npm version are you using?

test/integration/orca_serve_offline_test.js
```diff
@@ -0,0 +1,84 @@
...
+const Application = require('spectron').Application
+
+const { paths } = require('../common')
+
+const PORT = 9109 + 2
+
+const numberOfComponents = 6
+const axios = require('axios')
+const fs = require('fs')
+const pathToPlotlyJS = '/tmp/plotly-latest.min.js'
```
etpinard: Let's put that in `./build` aka [`paths.build`](https://github.com/plotly/orca/blob/2bee0c4333cf7327081f4852dde9c8b3c23e8e5c/test/common.js#L11) and use `path.join` so that your path is compatible on Windows:

```js
const = path.join(paths.build, 'plotly-latest.min.js')
```

test/integration/orca_serve_offline_test.js
```diff
@@ -0,0 +1,84 @@
+const tap = require('tap')
+const Application = require('spectron').Application
+
+const { paths } = require('../common')
+
+const PORT = 9109 + 2
```
etpinard: `9111`

test/integration/orca_serve_offline_test.js
```diff
@@ -0,0 +1,84 @@
...
+tap.test('should not link to resources on the network', t => {
+  var promises = []
+  for (var i = 0; i < numberOfComponents; i++) {
+    promises.push(getScriptTagsSrc(0))
+  }
+  Promise.all(promises).then(values => {
+    values.forEach(result => {
+      var urls = result.value
+      urls.forEach(url => {
+        t.notOk(url.match('http'), `A script tag refers to an HTTP resource ${url}`)
```
etpinard: Shouldn't we test `https` also?

test/integration/orca_serve_offline_test.js
antoinerg: Done in 3e107345b986a71601f8ff4cf845e7d9877463b1!

test/integration/orca_serve_offline_test.js
antoinerg: Brillant! Done in 3e107345b986a71601f8ff4cf845e7d9877463b1

test/integration/orca_serve_offline_test.js
antoinerg: I think this works since `https` will match `http`!

etpinard: Ah I see. Could you change the assert msg to

```
`A script tag refers to an HTTP(S) resource ${url}`
```

then to avoid any potential confusion.

package-lock.json
antoinerg: I'm using `6.4.1`... Sorry about that. Using `6.5.0` gives me `sha512` hashes as expected. 

What do you think of the idea of adding `"npm": "^6.5.0" ` in `devDependencies`?

test/integration/orca_serve_offline_test.js
antoinerg: Done in d29367d

package-lock.json
antoinerg: I didn't add `npm` in `devDependencies` but used a newer npm version to update `package-lock.json` in commit 78f3dd8a583cca869617aaee7bc1c56e69713e87.

Putting `npm` in `devDependencies` would make sense to me. That way we would all use the exact same version by running `./node_modules/.bin/npm`. We can discuss that later :)

etpinard: Hmm. You're suggesting listing `npm` as a dependency of things to install with `npm`? That's sounds strange. I've never seen that before. Maybe it's a thing though.

etpinard:
:dancer:

package-lock.json
antoinerg: > You're suggesting listing `npm` as a dependency of things to install with `npm`?

Said that way it sounds kinda stupid indeed :laughing:. We could instead update the `CONTRIBUTING.md` to state what version of `npm` is required.

chriddyp: We use the `engines` property in `package.json` for this in Dash:
https://github.com/plotly/dash-table/blob/03b59fd53cea0c61a42a3124bb126b24d29d51b3/package.json#L91-L94, https://docs.npmjs.com/files/package.json#engines

antoinerg: @etpinard, see @chriddyp's comment above. That's exactly the feature I was looking for: don't let me change `package-lock.json` if my npm version is wrong! Maybe we should add this here and in plotly.js' repo.

PR #192, opened 2019-01-18
ensure circleci doesn't overwrite Docker images tagged for release

antoinerg:
This PR prevents CircleCI from tagging Docker images with names reserved for releases (ie. of the form `x.x.xb.x` where `x` is one or multiple digits).


etpinard:
Looks good :dancer:

Issue #193, opened 2019-01-24
Scatter Box Map Plot Error 525


skuzmier:
When I try to save a map box scatter plot as a .png using pio.write_image(fig)
```
ValueError: 
The image request was rejected by the orca conversion utility
with the following error:
   525: plotly.js error
```
When I generate an interactive plot everything works fine and when I save the image manually it looks like this.

![image](https://user-images.githubusercontent.com/12565852/51658157-2726ce00-1f75-11e9-880f-b788dad7a08f.png)

The plot is created using the following function:
```
def plot_caiso_map(df, dt, col):
    #set the geo=spatial data
    if 'ldt' not in df.columns:
        df.reset_index(inplace=True)
    df.set_index('ldt', inplace=True)
    df['hover_info'] =  df['node'] +  '  |  ' +df[col].round(2).astype(str)
    df = df[dt[0]: dt[0]]
    data = [go.Scattermapbox(
            lat= df['lat'] ,
            lon= df['long'],
            customdata = df['node'],
            mode='markers',
            marker=dict(
                size= 6,
                color= df[col],
                colorscale= 'Jet',
                showscale=True,
                cmax= min(200, max(df[col].max(), 100)),
                cmin= max(-40, min(df[col].min(), -20)),
                opacity = .8,
            ),
        #text = df['hover_info'],
          )]
    #set the layout to plot
    layout = go.Layout(autosize=False,
                   mapbox= dict(accesstoken='MY_TOKEN',
                                bearing=0,
                                pitch=0,
                                zoom=5,
                                center= dict(lat=36.95,
                                             lon=-117.0),
                                style='dark'),
                    width=900,
                    height=700, 
                    title = "CAISO %s - %s" %(col.upper(), dt))
    df.reset_index(inplace=True)
    fig =  go.FigureWidget(data=data, 
                     layout=layout
                    )
    pio.write_image(fig, 'prices_map.png')
    py.iplot(fig)
    return fig

```
This issue may be related to issue #188  but the error messages were different enough that I though this might be seperate. 

If I try saving the figure to JSON and then running from command line with:

orca graph figure.json --mapbox-access-token MY_ACCESS_TOKEN  -o figure.png --verbose
I get the following error message:

```
gpu process crashed - false
          Chrome version 59.0.3071.115
          Electron version 1.8.4
gpu process crashed - false
          Chrome version 59.0.3071.115
          Electron version 1.8.4
gpu process crashed - false
          Chrome version 59.0.3071.115
[...]

PR #195, opened 2019-01-28
Pass mapboxAccessToken as null to Plotly.toImage when not set

etpinard:
... so that it does not override `layout.mapbox.accesstoken` - fixes https://github.com/plotly/orca/issues/193 cc @antoinerg


antoinerg:
Thanks for the quick fix @etpinard 

:dancer:

PR #197, opened 2019-01-30
Remove broken key servers from Dockerfile

### LINKED ISSUE #196: Orca Dockerfile wastes time on broken keyservers
Orca's current Dockerfile tries and fails to fetch each needed GPG key from the known broken keyservers `pgp.mit.edu` and `keyserver.pgp.com` before falling back to the reliable high-availability pool `ha.pool.sks-keyservers.net`. This causes Docker image builds to take far too long, and frequently makes the build jobs fail entirely.

We should remove the two broken keyservers from the Dockerfile. PR forthcoming for @antoinerg to review as soon as I'm done testing the change.

FYI @plotly/devops - once the corresponding PR is merged, any of you doing orca work should be sure to update your orca branches for an easier time with CircleCI.

jkaplowitz:
Fixes #196 

FYI @plotly/devops. @antoinerg please review.

### DIFF
--- a/deployment/Dockerfile
+++ b/deployment/Dockerfile
@@ -23,8 +23,6 @@ RUN set -ex \
     C4F0DFFF4E8C1A8236409D08E73BC641CC11F4C8 \
     56730D5401028683275BD23C23EFEFE93C4CFFFE \
   ; do \
-    gpg --keyserver pgp.mit.edu --recv-keys "$key" || \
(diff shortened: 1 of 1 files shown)


antoinerg:
Thanks @jkaplowitz for the nice :hocho: ! 

:dancer: once `docker-build-and-push` succeeds :heavy_check_mark:

PR #198, opened 2019-01-30
Ignore certificate errors if indicated by env var

jkaplowitz:
Relates to plotly/streambed#11906. I've tested this and confirmed that certificate errors are ignored / PDFs can be rendered from self-signed websites if the environment variable is set as indicated. No change without the environment variable.

FYI @plotly/devops. @etpinard please review for team orca; @joaoalf please review for the devops team.

### DIFF
--- a/deployment/run_server
+++ b/deployment/run_server
@@ -6,6 +6,10 @@ if [[ $PLOTLY_CUSTOM_FONTS_ENABLED = true ]]; then
   fc-cache -v /usr/share/fonts/customer
 fi
 
+if [[ $PLOTLY_IMAGESERVER_IGNORE_CERT_ERRORS = true ]]; then
+  ORCA_IGNORECERTERRORS_ARG="--ignore-certificate-errors"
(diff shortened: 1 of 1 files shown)


deployment/run_server
```diff
@@ -6,6 +6,10 @@ if [[ $PLOTLY_CUSTOM_FONTS_ENABLED = true ]]; then
   fc-cache -v /usr/share/fonts/customer
 fi
 
+if [[ $PLOTLY_IMAGESERVER_IGNORE_CERT_ERRORS = true ]]; then
+  ORCA_IGNORECERTERRORS_ARG="--ignore-certificate-errors"
```
antoinerg: For reference, flag [`--ignore-certificate-errors` is documented here in Electron 1.8](https://github.com/electron/electron/blob/1-8-x/docs/api/chrome-command-line-switches.md#--ignore-certificate-errors) and is available in newer versions of Electron as well!

PR #206, opened 2019-02-12
"v1.0-onprem" merge to master

scjody:
This PR contains fixes that were made on `v1.0-onprem` but not `master`. They are needed for the upcoming 3.0.3e On-Prem release to address https://github.com/plotly/streambed/issues/11780


etpinard:
:dancer: when tests pass (you might need to restart `docker-build-and-push` which has been flaky)

PR #207, opened 2019-02-12
V1.0-onprem merge to 3.0-release

scjody:
This PR contains fixes that were made on `v1.0-onprem` but not `master`. I would like to put them in a 3.0.x maintenance release to fix plotly/streambed#11780.


etpinard:
ok looks good :dancer:

PR #208, opened 2019-02-13
Orca startup time workaround

scjody:
Workaround for https://github.com/plotly/streambed/issues/11780, plus debugging improvement.

* [x] Tested with CI-built image

@jkaplowitz Please review.

### DIFF
--- a/deployment/monitrc
+++ b/deployment/monitrc
@@ -2,8 +2,8 @@ set daemon 10
 set init
 
 check process xvfb with pidfile "/var/run/xvfb.pid"
-  start program = "/usr/local/bin/xvfb_wrapper"
-  restart program = "/usr/local/bin/xvfb_wrapper"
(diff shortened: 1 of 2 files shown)


scjody:
> I think the script's fd configuration gets shared by child processes unless/until they take action to redirect theirs.

This is correct, and this is why we don't need the explicit redirection on the `xvfb-run ...` line anymore.

> The redirections on new line 26 differ from those on former line 38 in one way: if pid 1 has different destinations for its own stdout and stderr, the old code sends the script's stdout and stderr to the corresponding pid 1 destinations, whereas the new code sends them both to pid 1's stdout.

You're right about that, and I didn't intend to make that change. I've changed it to keep stdout and stderr separate, just in case someone is relying on this behaviour. (`docker logs` does keep them separate.)

jkaplowitz:
Great. The new revision looks good. :dancer:

Issue #209, opened 2019-02-14
Intermittent Orca build failure due to --recv-keys failure


scjody:
https://github.com/plotly/orca/pull/197 attempted to fix the Docker build run by CI (`docker-build-and-push` step) but the build still fails frequently (>75% of the time) with:

```
Step 4/23 : RUN set -ex   && for key in     9554F04D7259F04124DE6B476D5A82AC7E37093B     94AE36675C464D64BAFA68DD7434390BDBE9B9C5     FD3A5288F042B6850C66B31F09FE44734EB7990E     71DCFD284A79C3B38668286BC97EC7A07EDE3FC1     DD8F2338BAE7501E3DD5AC78C273792F7D83545D     B9AE9905FFD7803F25714661B63B535A4C206CA9     C4F0DFFF4E8C1A8236409D08E73BC641CC11F4C8     56730D5401028683275BD23C23EFEFE93C4CFFFE   ; do     gpg --keyserver ha.pool.sks-keyservers.net --recv-keys "$key" ;   done
 ---> Running in 73e9550ab7db
+ gpg --keyserver ha.pool.sks-keyservers.net --recv-keys 9554F04D7259F04124DE6B476D5A82AC7E37093B
gpg: directory `/root/.gnupg' created
gpg: new configuration file `/root/.gnupg/gpg.conf' created
gpg: WARNING: options in `/root/.gnupg/gpg.conf' are not yet active during this run
gpg: keyring `/root/.gnupg/secring.gpg' created
gpg: keyring `/root/.gnupg/pubring.gpg' created
gpg: requesting key 7E37093B from hkp server ha.pool.sks-keyservers.net
gpgkeys: key 9554F04D7259F04124DE6B476D5A82AC7E37093B can't be retrieved
gpg: no valid OpenPGP data found.
gpg: Total number processed: 0
gpg: keyserver communications error: keyserver helper general error
gpg: keyserver communications error: unknown pubkey algorithm
gpg: keyserver receive failed: unknown pubkey algorithm
```

When it fails it generally fails on the key shown above (the first one we attempt to retrieve), but I've seen it fail on other keys about 10% of the time.

@plotly/devops @antoinerg @etpinard FYI

antoinerg:
The only reason we're downloading those keys is to install `nodejs`. Maybe we could simply start with a Docker image already containing `nodejs`.

@plotly/devops Is it mandatory to base our image from `ubuntu:xenial`?

scjody:
[...]

PR #210, opened 2019-02-14
Update keyservers used for nodejs installation

scjody:
The new commands come from the current node Dockerfile for Debian Stretch: https://github.com/nodejs/docker-node/blob/daa131d713cf42ae181292471766879f750b5230/10/stretch/Dockerfile

Fixes https://github.com/plotly/orca/issues/209

This passed CI 5 out of 5 times.

@antoinerg I'm still in favour of the updates you're proposing if you see value in them, but this is a much smaller change that I'm comfortable putting onto the `3.0-release` branch.

@jkaplowitz Please review.

### DIFF
--- a/deployment/Dockerfile
+++ b/deployment/Dockerfile
@@ -23,7 +23,9 @@ RUN set -ex \
     C4F0DFFF4E8C1A8236409D08E73BC641CC11F4C8 \
     56730D5401028683275BD23C23EFEFE93C4CFFFE \
   ; do \
-    gpg --keyserver ha.pool.sks-keyservers.net --recv-keys "$key" ; \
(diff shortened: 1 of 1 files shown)


antoinerg:
@scjody Since your team is in charge of deploying/testing this container in production, it should be your decision and we should do what is more comfortable for you.

On a related note, several users are interested in us releasing an official Docker image for Orca https://github.com/plotly/orca/issues/187 so I will revisit this `Dockerfile` in the near future and either propose updates or create a new `Dockerfile` for that purpose. To be followed! In the meantime, you can merge this in :dancer:

scjody:
@antoinerg I just commented in that issue. I don't see any reason not to advertise our existing quay.io repo.

PR #220, opened 2019-03-26
Docker: pin Inkscape's version and enable all EMF fixes

### LINKED ISSUE #218: EMF - problems with colorscale
A user reported that the color scale doesn't render properly in an EMF file produced by Orca:
fig emf

More details about how this result was obtained will follow.

antoinerg:
Closes #218 by enabling all EMF fixes available in Inkscape.

### DIFF
--- a/deployment/Dockerfile
+++ b/deployment/Dockerfile
@@ -113,8 +113,11 @@ RUN cd /opt && \
 # Install latest stable Inkscape
 RUN apt-get update && apt-get install -y software-properties-common python-software-properties \
     && add-apt-repository -y ppa:inkscape.dev/stable \
-    && apt-get update && apt-get install -y inkscape \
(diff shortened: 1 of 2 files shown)


etpinard:
Wow. Great fix! :dancer:

antoinerg:
For future reference, the following gist contains a script that `docker pull`, `run` and produce a EMF figure using the Docker image built from commit cb9c1e445596b447712806043d7a0ad966b39e9d:
```
git clone https://gist.github.com/272c1cbf8db24e4c2aceffc91d137bf7.git
cd 272c1cbf8db24e4c2aceffc91d137bf7
./render.sh
```
Then inspect the resulting `fig.emf` or `fig.emf.png`. We see that the colorbar renders properly:
fig emf

The above test should be automated and part of our test suite whenever we add image tests to Orca :tada:

PR #221, opened 2019-03-26
Merge EMF fix from `3.1-release` into `master`

antoinerg:
Merging `3.1-release` to `master` so the public has access to the EMF fixes. The commits have already been reviewed so just need a rubber stamp

Closes https://github.com/plotly/orca/issues/218

See original PR https://github.com/plotly/orca/pull/220


etpinard:
:dancer: !

PR #222, opened 2019-03-27
Release official Docker image + pixel exact image tests

### LINKED ISSUE #187: Add Docker in installation methods

antoinerg:
Closes #187 

@scjody suggested we reuse our existing `Dockerfile` to publish our official Docker images. In order to do so, I had to adapt its `ENTRYPOINT` to support CLI invocation (see https://github.com/plotly/orca/issues/187#issuecomment-471095473 for details). If the first argument is either `--version`, `--help` or `graph` we assume Orca is used as a CLI. If the first argument is `serve` or anything else, we launch the server as we used to do.

- [x] Publish Docker image as `plotly/orca`
- [x] Support CLI invocation
- [x] Advertise our Docker image in the documentation
- [x] Add image tests

## image tests

The inclusion of image tests in this PR eliminates the risk of missing dependencies in the final Docker image. By using software rendering, we get reproducible images that we can compare pixel per pixel with a set of baselines in an exact manner. By visually testing all image formats (`pdf`, `eps`, `svg`, `png`, `emf`), we catch missing system dependencies such as Poppler, Inkscape, Mathjax or missing fonts. Hopefully, this will catch regressions whenever we perform system upgrades.

- [x] document image tests in CONTRIBUTING.md
- [x] runs automatically on CircleCI as a final job

@plotly/devops please have a look at the new entrypoint.


jonmmease:
Thanks for working on this @antoinerg, this is going to be really helpful!

I'd like to see if we can work through how Python/R/Julia users could take advantage of orca as a docker image without each library needing to include docker related logic.  Would it be possible to distribute a shell script, `orca.sh`, that presents that same interface as the native orca binary?  Something like 

if `orca graph` then call `docker run -i quay.io/plotly/orca graph "$@"`

if `orca serve --port=8123` then call `docker run -d -p 8123:8123 quay.io/plotly/orca "$@"`

The R and Julia libraries use the `orca graph` command line API, while plotly.py uses the `orca serve` API with `--port` and `--graph-only` options.  If we had a shell script like this for users to put on their path then I think all of the client libraries would work without modification.  Can `docker pull` add something to the PATH or would that have to be a manual step?

What do you think?

antoinerg:
> Thanks for working on this @antoinerg, this is going to be really helpful!

I'm glad to see lots of people are excited about this one!

> I'd like to see if we can work through how Python/R/Julia users could take advantage of orca as a docker image without each library needing to include docker related logic. Would it be possible to distribute a shell script, orca.sh, that presents that same interface as the native orca binary?

That is certainly one of the goal here. Making an `orca.sh` shell script calling Docker as you suggested would be great and I will try to add one to this PR.

> Can docker pull add something to the PATH or would that have to be a manual step?

As far as I know, it cannot. It would have to be a manual step.

scjody:
Sounds like a great idea to me!

Minor request: if you're adding a tool that users are expected to run, it should not have a `.sh` (or `.bash`, `.py`, etc.) extension. [Rationale](https://google.github.io/styleguide/shell.xml?showone=File_Extensions#File_Extensions)

Also note that we have a company-wide [bash style guide](https://github.com/plotly/dev-docs/blob/master/reference/styleguides/bash.md) (mostly incorporates Progrium's).

deployment/Dockerfile
```diff
@@ -79,10 +79,8 @@ RUN apt-get update -y && \
 ####################
 # Download mathjax (same version as plotly.js extras/)
 
-RUN apt-get update -y && apt-get install -y subversion && \
-    rm -rf /var/lib/apt/lists/* && apt-get clean && \
-    svn checkout https://github.com/plotly/plotly.js.git/trunk/dist/extras/mathjax && \
-    apt-get --auto-remove -y remove subversion
+RUN curl -L https://github.com/plotly/plotly.js/archive/master.tar.gz \
+    | tar -xvzf - --strip-components=3 plotly.js-master/dist/extras/mathjax
```
antoinerg: Replacing `svn` by `git` is a major win because the `svn checkout` from Github often failed!

.gitignore
```diff
@@ -10,3 +10,6 @@ coverage
 deployment/*.retry
 
 /yarn-error.log
+
+test/image/build
```
etpinard: Could you use 

https://github.com/plotly/orca/tree/master/build

instead, like our other tests?

deployment/Dockerfile
etpinard: Awesome :tada:

.gitignore
antoinerg: Done in 697068a

antoinerg:
@jonmmease I wrote a little shell script in commit 94bb0a4. Please see the commit message for details. Also, can you test it out on MacOS and adapt it if it doesn't work. Anyway, let me know what you think!

.gitignore
```diff
@@ -10,3 +10,6 @@ coverage
...
+test/image/diff
```
etpinard: `test/image/build` and `test/image/diff` aren't used anymore after https://github.com/plotly/orca/commit/697068a13a470c9b74d644f255fb2f72eeeddeed , correct?

test/image/render_mocks_cli
```diff
@@ -0,0 +1,44 @@
...
+do
+  # To obtain reproducible results, we need to use software rendering
+  export LIBGL_ALWAYS_SOFTWARE=true
+  export GALLIUM_DRIVER=softpipe
+
+  # Supply defaults assets to Orca
+  ORCA_OPTS=(--topojson /plotly-geo-assets.js --mathjax /mathjax/MathJax.js)
+
+  # pin Plotly.js' version
+  ORCA_OPTS+=(--plotlyJS "https://cdn.plot.ly/plotly-1.47.2.min.js")
```
etpinard: Hmm. Does the soon-to-be public docker container use `plotly-latest.min.js` by default? Or a tagged version?

test/image/render_mocks_cli
```diff
@@ -0,0 +1,44 @@
...
+
+  # Set Orca options
+  ORCA_OPTS+=(--verbose --format "$format" --output-dir "$1")
+
+  if [ -z "$2" ]; #
+  then
+    echo "Executing locally"
```
etpinard: Could you log info about the format tested?

The log currently look like:

image

so maybe having _Executing locally. Format $format_ would be an improvement.

.gitignore
antoinerg: Correct! Reverted `.gitignore` in 36bb04f6275617737b2ce21645266ce0e53eb300

test/image/render_mocks_cli
etpinard: image

I was expecting a slower job. Nicely done!

test/image/render_mocks_cli
antoinerg: Yes it will use whatever Orca uses by default which is `plotly-latest.min.js`.

The reason we specify it here is to make sure we use the same version of plotly.js used to generate the baselines.

.gitignore
etpinard: :tada:

test/image/render_mocks_cli
etpinard: OK, that's great.

Could you add some info under https://github.com/plotly/orca/blob/docker-release/CONTRIBUTING.md#running-tests about the image tests and where to update the plotly.js version used to generate the baselines when needed.

test/image/render_mocks_cli
antoinerg: Done in 3a37405b6932fd0d5659ebfdd6a1080ba4914580

etpinard: :tada:

test/image/render_mocks_cli
antoinerg: Done in f5ce6f9b4231d930c699d635ec38fd8b4785a9f7

etpinard: Fantastic :tada:

antoinerg:
Any comments on this PR? @etpinard @jonmmease @plotly/devops 

Thank you all!

etpinard:
:dancer: from me

PR #228, opened 2019-04-15
implement http content-negotiation for plotly-graph

### LINKED ISSUE #226: Add support for HTTP "Content negotiation"
I think it would be nice to support [content-negotiation](https://en.wikipedia.org/wiki/Content_negotiation) so a client can specify which format he wants via HTTP header `Accept` without having to modify the payload:
```
POST / HTTP/1.1
Host: orca
User-Agent: curl/7.61.0
Accept: image/svg+xml

$FIGURE_JSON_IN_BODY
```

antoinerg:
Closes #226 

Check added tests for details!

- [x] Return default format when HTTP `accept: */*` and `accept: *`


test/integration/orca_serve_test.js
```diff
@@ -63,6 +63,70 @@ tap.test('should work for *plotly-graph* component', t => {
...
+    }, (err, res, body) => {
+      if (err) t.fail(err)
+
+      t.equal(res.statusCode, 200, 'code')
+      t.equal(res.headers['content-type'], 'image/svg+xml')
+      t.end()
+    })
+  })
+
+  t.test('it should override format provided in payload', t => {
```
antoinerg: I'm not sure about this one... :thinking: 

Problematic case: the user asks for SVG in HTTP header but also ask for PNG in the payload via `format` attribute. Which one should win?

src/app/server/create-server.js
```diff
@@ -157,7 +157,7 @@ function createServer (app, BrowserWindow, ipcMain, opts) {
       }
 
       pending++
-      comp._module.parse(body, compOpts, sendToRenderer)
+      comp._module.parse(body, compOpts, sendToRenderer, req)
```
etpinard: Hmm. I would prefer leaving `sendToRenderer` as the last argument, as callbacks `sendToRenderer` as usually passed last. 

I think having 

```js
 comp._module.parse(body, req, compOpts, sendToRenderer)
```

would be best ... and don't forget to update:

https://github.com/plotly/orca/blob/master/CONTRIBUTING.md#nomenclature-for-ipc-callbacks

accordingly.

test/integration/orca_serve_test.js
etpinard: I think you have it the right way here: header overriding payload.

antoinerg: @etpinard I decided to go with payload overriding HTTP header in 9d74f00

The reason is that I don't want to introduce a breaking change in the HTTP API. So for users that defines the format in their payload, Orca should behave the same way.

src/app/server/create-server.js
antoinerg: Done in c1e0ab5

bin/orca.js
```diff
@@ -2,7 +2,7 @@
 
 const path = require('path')
 const { spawn } = require('child_process')
-const electronPath = require('electron')
+const electronPath = process.env['ELECTRON_PATH'] ? process.env['ELECTRON_PATH'] : require('electron')
```
etpinard: Why do we need this?

antoinerg: It is not mandatory for this PR but it provides a way to use a manually installed `electron` in the event that [electron-download](https://github.com/electron-userland/electron-download) fails to install pre-built binary for your platform.

I could remove it but it's a must-have in my current dev environment and maybe it also the case for others. It simply gives more flexibility. What do you think?

test/unit/server_test.js
```diff
@@ -184,6 +184,34 @@ tap.test('createServer:', t => {
     })
   })
 
+  t.test('it should do HTTP content-negotation for *plotly-graph* component', t => {
```
etpinard: What happened to your other content-negotiation tests? Don't you want to test a valid content negotiation?

antoinerg: Sorry for the oversight! Thank you so much for the spotting this @etpinard. Removing the tests in my previous commit was a mistake.

Fixed in a95efcf

etpinard:
Nicely done :dancer:

Issue #164, opened 2018-12-04
ConnectionRefused on first attempt at exporting image


dwjp:
I'm running orca in a miniconda container, which is in turn running as a celery service (though probably not relevant).

On the first one or two attempts at generating an image (when kicking off at about the same time, but on separate processes), I get connection refused. On subsequent tries, it works fine, everytime. 

At a guess, I suspect the python client isn't waiting long enough for orca to start, but makes it through on a warm restart when orca starts a bit quicker?

>   File "/opt/conda/lib/python2.7/site-packages/plotly/io/_orca.py", line 1298, in to_image
>     height=height)
>   File "/opt/conda/lib/python2.7/site-packages/retrying.py", line 49, in wrapped_f
>     return Retrying(*dargs, **dkw).call(f, *args, **kw)
>   File "/opt/conda/lib/python2.7/site-packages/retrying.py", line 212, in call
>     raise attempt.get()
>   File "/opt/conda/lib/python2.7/site-packages/retrying.py", line 247, in get
>     six.reraise(self.value[0], self.value[1], self.value[2])
>   File "/opt/conda/lib/python2.7/site-packages/retrying.py", line 200, in call
>     attempt = Attempt(fn(*args, **kwargs), attempt_number, False)
>   File "/opt/conda/lib/python2.7/site-packages/plotly/io/_orca.py", line 1205, in request_image_with_retrying
>     response = requests.post(server_url + '/', data=json_str)
>   File "/opt/conda/lib/python2.7/site-packages/requests/api.py", line 112, in post
>     return request('post', url, data=data, json=json, **kwargs)
>   File "/opt/conda/lib/python2.7/site-packages/requests/api.py", line 58, in request
>     return session.request(method=method, url=url, **kwargs)
>   File "/opt/conda/lib/python2.7/site-packages/requests/sessions.py", line 518, in request
>     resp = self.send(prep, **send_kwargs)
>   File "/opt/conda/lib/python2.7/site-packages/requests/sessions.py", line 639, in send
>     r = adapter.send(request, **kwargs)
>   File "/opt/conda/lib/python2.7/site-packages/requests/adapters.py", line 502, in send
>     raise ConnectionError(e, request=request)
[...]

PR #237, opened 2019-07-03
bump electron to 4.2.6

antoinerg:
Although most SVG baselines stayed the same, there are slight differences in rendering for the other baselines (most certainly due to switching to a much newer version of Chromium!).

This PR essentially supersedes https://github.com/plotly/orca/pull/125

- [ ] Test that this PR closes https://github.com/plotly/orca/issues/124


etpinard:
Thanks very much for taking this on @antoinerg !!

- [ ] We should also make sure @plotly/devops tests the new docker build on stage

etpinard:
> * Test that this PR closes #124

cc @jonmmease here

scjody:
> We should also make sure @plotly/devops team tests the new docker build on stage

Normally the developer making the change should be able to test it. `@plotbot deploy bump-electron-4 to imageserver stage` should get your changes onto stage.

If you need us to test this please let us know in detail what needs to be tested. It looks to me like the fix only affects Python users so do you just need a general test that Orca still works?

etpinard:
> Normally the developer making the change should be able to test it.

I agree, this would be ideal. But, as far as I know, there are no docs on how to test things like: CPU/memory usage, startup time and other things that might be of interest for the image server.

scjody:
> I agree, this would be ideal. But, as far as I know, there are no docs on how to test things like: CPU/memory usage, startup time and other things that might be of interest for the image server.

OK, I didn't realize that was what you wanted. Someone from the team can do that this time and write some documentation for next time.

For CPU and memory usage, I suggest we send the set of test images to the Orca staging server and measure what happens. Do you have a good way to do this? Last time I needed to do this I (wrote a tool)[https://github.com/plotly/streambed/issues/9865#issuecomment-363225820], but it needs some work if it's something we want to keep using.

Do you have any specific concerns other than CPU/memory usage and startup time?

antoinerg:
@scjody in the spirit of not wasting your team's time, let's wait before testing: I will try to bump Ubuntu's version as well.

tonkolviktor:
Hello,

i was having troubles with 1.2.1 version of orca on my raspberry pi.

orca tag v3.2.0 did not work as well: i was getting this error: https://github.com/plotly/orca/issues/164
orca tag v3.20 + this PR did not work as well: because of this: https://github.com/electron/electron/issues/16205
orca tag v3.20 + electron 3.1.3 worked (on raspberry)!

antoinerg:
> orca tag v3.20 + electron 3.1.3 worked (on raspberry)!

@tonkolviktor you might be interested to know specifically set the `electron` binary used by Orca by setting the env variable `ELECTRON_PATH`:
https://github.com/plotly/orca/blob/f40962271ce901492513115854cceb45977448eb/bin/orca.js#L5

antoinerg:
Closing in favor of PR https://github.com/plotly/orca/pull/266

PR #243, opened 2019-09-06
Fix custom pageSize for Dash App Previews

tarzzz:
Part of https://github.com/plotly/streambed/issues/13458

  - [x] Test different size combinations in an on-prem environment.

The custom size is not working because `remote.CreateBrowserWindow` expects the height/width to be "non-strings" (ie integers or floats). This fixes the issue.

### DIFF
--- a/src/component/plotly-dash-preview/parse.js
+++ b/src/component/plotly-dash-preview/parse.js
@@ -45,8 +45,8 @@ function parse (body, req, opts, sendToRenderer) {
       height: body.pageSize.height * cst.pixelsInMicron
     }
     result.pdfOptions.pageSize = {
-      width: body.pageSize.width,
-      height: body.pageSize.height
+      width: Math.ceil(body.pageSize.width),
+      height: Math.ceil(body.pageSize.height)
     }
   } else {
     return errorOut(
--- a/src/component/plotly-dash-preview/parse.js
+++ b/src/component/plotly-dash-preview/parse.js
@@ -44,6 +44,10 @@ function parse (body, req, opts, sendToRenderer) {
       width: body.pageSize.width * cst.pixelsInMicron,
       height: body.pageSize.height * cst.pixelsInMicron
     }
+    result.pdfOptions.pageSize = {
+      width: body.pageSize.width,
+      height: body.pageSize.height
+    }
   } else {
     return errorOut(
       400,
--- a/test/unit/plotly-dash-preview_test.js
+++ b/test/unit/plotly-dash-preview_test.js
@@ -54,6 +54,10 @@ tap.test('parse:', t => {
         height: 3.779527559055118,
         width: 3.779527559055118
       })
+      t.same(result.pdfOptions.pageSize, {
+        height: 1000,
+        width: 1000
+      })
       t.end()
     })
   })


src/component/plotly-dash-preview/parse.js
```diff
@@ -41,8 +41,8 @@ function parse (body, req, opts, sendToRenderer) {
   } else if (body.pageSize && isPositiveNumeric(body.pageSize.width) &&
              isPositiveNumeric(body.pageSize.height)) {
     result.browserSize = {
-      width: body.pageSize.width * cst.pixelsInMicron,
-      height: body.pageSize.height * cst.pixelsInMicron
+      width: parseFloat(body.pageSize.width * cst.pixelsInMicron),
```
antoinerg: Can you change the commit message to `Ensure that width/height settings are Floats`

antoinerg:
Non-blocking comment https://github.com/plotly/orca/pull/243/files#r321845954

Other than that, feel free to merge :dancer:

src/component/plotly-dash-preview/parse.js
etpinard: I think

```js
width: Number(body.pageSize.width) * cst.pixelsInMicron
```

would make things more clear.

etpinard: Not that it matters much here, but `parseFloat` is slow.

src/component/plotly-dash-preview/parse.js
```diff
@@ -40,10 +40,10 @@ function parse (body, req, opts, sendToRenderer) {
     result.browserSize = cst.sizeMapping[result.pdfOptions.pageSize]
   } else if (body.pageSize && isPositiveNumeric(body.pageSize.width) &&
              isPositiveNumeric(body.pageSize.height)) {
-    result.browserSize = {
-      width: body.pageSize.width * cst.pixelsInMicron,
-      height: body.pageSize.height * cst.pixelsInMicron
-    }
+    result.pdfOptions.pageSize = {
+      width: body.pageSize.width, 
+      height: body.pageSize.height
```
antoinerg: This is a breaking change :\

antoinerg: @tarzzz I don't think we should make such a big change to the units for `widht` and `height`. Is it not possible to update your client instead of Orca?

chriddyp:
I think only integers are supported - there's an electron bug with floats: https://github.com/electron/electron/issues/9361#issue-225991341

tarzzz:
In my investigations, I noticed that the `pageSize` specified in the request body is not being passed to `printToPDF` call.  This fixes the issue.

> I think only integers are supported - there's an electron bug with floats:https://github.com/electron/electron/issues/9361#issue-225991341

We can `Math.ceil` the incoming values to ensure only integers are passed along. Choosing b/w `Math.ceil` or `Math.floor` or `Math.round` does not matter because it will only change by +- 1 micron.

tarzzz:
@antoinerg or @etpinard Please review.. I have checked this in an on-prem environment and custom page sizes work for me.

joaoalf:
@tarzzz Don't we want this for `3.2-release`?

etpinard:
Looks good to me :dancer: 

That said, are folks ok with providing the page size in microns? Maybe we could change it to pixels, call this a breaking change and bump `orca` to v2?

tarzzz:
> are folks ok with providing the page size in microns

It is not a breaking change as the `pageSize` has always been in microns in the API.

etpinard:
Correct @tarzzz your PR here does not introduce a breaking change.

But _what if_ we'd like to improve the API (here OR in a future PR), I think making users input page dimensions in pixels would be beneficial.

tarzzz:
> But what if we'd like to improve the API (here OR in a future PR), I think making users input page dimensions in pixels would be beneficial.

Yes, but this is going to be the part of a maintenance release.. so we should not be introducing breaking change(s) or new features. 

If required, We can add the new issue (for dimensions in pixels) as the part of the next release work.

antoinerg:
@tarzzz thanks again for taking on this issue. Could you also make a PR to merge those changes into `master`?

PR #247, opened 2019-09-13
Add differentiation from orca screen reader

cheog:
There's a package called orca that's a screen reader. I didn't know I had it installed and it took me about three to four hours to figure it out because there is no mention of this on stack overflow or the community forums (at least that I could find). So this simply adds a note about that in the readme. If this is rejected I'd be happy to make and answer a post about this on stackoverflow.

PS this is my first pull request ever so if I'm doing something wrong please tell me.

### DIFF
--- a/README.md
+++ b/README.md
@@ -151,6 +151,13 @@ These additional dependencies can be satisfied by installing:
  - The `google-chrome-stable` package from the [Google Linux Software
  Repository](https://www.google.com/linuxrepositories/).
 
+##### Linux Troubleshooting: Server is started but plotly can't connect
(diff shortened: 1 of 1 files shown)


antoinerg:
Thanks @cheog for the PR! I think however that this troubleshooting tip would be better suited for the documentation of [plotly.py](https://github.com/plotly/plotly.py). I will therefore close this PR.

ahmedhosny:
+1 - I faced the same problem and highly suggest updating the docs. For those on linux (and not using conda), installing orca requires some effort and anything to make that easier is much appreciated!

PR #250, opened 2019-09-23
/dash-preview: Fix window size

tarzzz:
Detailed explanation in: https://github.com/plotly/streambed/issues/13458#issuecomment-534274326

### DIFF
--- a/deployment/run_server
+++ b/deployment/run_server
@@ -40,7 +40,7 @@ pkill -9 Xvfb
 pkill -9 node
 pkill -9 electron
 
-xvfb-run --auto-servernum --server-args '-screen 0 640x480x24' /var/www/image-exporter/bin/orca.js serve $REQUEST_LIMIT --safe-mode --verbose $PLOTLYJS_ARG $ORCA_IGNORECERTERRORS_ARG $@ &
+xvfb-run --auto-servernum --server-args '-screen 0 1024x768x24' /var/www/image-exporter/bin/orca.js serve $REQUEST_LIMIT --safe-mode --verbose $PLOTLYJS_ARG $ORCA_IGNORECERTERRORS_ARG $@ &
 echo \$! > \$PIDFILE
 
 EOF
--- a/src/component/plotly-dash-preview/render.js
+++ b/src/component/plotly-dash-preview/render.js
@@ -15,6 +15,8 @@ function render (info, opts, sendToMain) {
   const result = {}
 
   let createBrowserWindowOpts = info.browserSize ? info.browserSize : {}
+  createBrowserWindowOpts['enableLargerThanScreen'] = true
+  createBrowserWindowOpts['useContentSize'] = true
   createBrowserWindowOpts['show'] = opts.debug
 
   let win = remote.createBrowserWindow(createBrowserWindowOpts)
--- a/test/unit/plotly-dash-preview_test.js
+++ b/test/unit/plotly-dash-preview_test.js
@@ -51,8 +51,8 @@ tap.test('parse:', t => {
 
       // height/width are converted from microns to pixels:
       t.same(result.browserSize, {
-        height: 3.779527559055118,
-        width: 3.779527559055118
+        height: 4,
+        width: 4
       })
       t.same(result.pdfOptions.pageSize, {
         height: 1000,
--- a/test/unit/plotly-dash-preview_test.js
+++ b/test/unit/plotly-dash-preview_test.js
@@ -6,10 +6,6 @@ const constants = require('../../src/component/plotly-dash-preview/constants')
 const remote = require('../../src/util/remote')
 const { createMockWindow } = require('../common')
 
-function clone (obj) {
-  return JSON.parse(JSON.stringify(obj))
-}
-
 tap.test('parse:', t => {
   const fn = _module.parse
 
@@ -36,16 +32,13 @@ tap.test('parse:', t => {
     })
   })
   t.test('pageSize options:', t => {
-    const mock = {
+    const mock = () => ({
       url: 'https://dash-app.com',
       selector: 'dummy'
-    }
+    })
 
     t.test('should error when not given', t => {
-      fn({
-        url: 'https://dash-app.com',
-        selector: 'dummy'
-      }, {}, {}, (errorCode, result) => {
+      fn(mock(), {}, {}, (errorCode, result) => {
         t.equal(errorCode, 400)
         t.same(result.msg, 'pageSize must either be A3, A4, A5, Legal, Letter, ' +
                            'Tabloid or an Object containing height and width in microns.')
@@ -86,7 +79,7 @@ tap.test('parse:', t => {
       var toplevel = arg[0]
       var pageSize = arg[1]
       t.test(`should size window and page properly when ${toplevel ? '' : 'pdf_options.'}pageSize is given`, t => {
-        var body = clone(mock)
+        var body = mock()
         if (toplevel) {
           body.pageSize = pageSize
         } else {
@@ -102,7 +95,7 @@ tap.test('parse:', t => {
     })
 
     t.test('should passthrough pdf_options', t => {
-      var body = clone(mock)
+      var body = mock()
       body.pdf_options = { pageSize: 'Letter', marginsType: 1, crazyOptions: true }
       fn(body, {}, {}, (errorCode, result) => {
         t.equal(errorCode, null)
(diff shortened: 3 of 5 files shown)


tarzzz:
@antoinerg or @etpinard Please review.

src/component/plotly-dash-preview/render.js
```diff
@@ -14,16 +14,19 @@ const cst = require('./constants')
...
 
-  let win = remote.createBrowserWindow(createBrowserWindowOpts)
+  let win = remote.createBrowserWindow(windowOpts)
   const contents = win.webContents
   const session = contents.session
 
   // Clear cookies before loading URL
   session.clearStorageData({}, () => {
     win.loadURL(info.url)
+    win.webContents.executeJavaScript(`window.resizeTo(${windowOpts.width}, ${windowOpts.height})`)
```
antoinerg: Why is this necessary? It seems like you already set the window's size. :thinking:

tarzzz: For some reason, it does not work when you do it with :```let win = remote.createBrowserWindow(createBrowserWindowOpts)```
It seems to be doing nothing.

I also tried `win.setSize` and `win.setContentSize` but they give me an error.

tarzzz: Actually, I think it is because we need to round off window size as well:
image

I will try that and get back to you..

antoinerg:
> y/streambed#13458 (comment)

If I understand correctly, the idea is to have the size of the viewport match the size of the page so that when printing to PDF, figures are already appropriately sized.

For this to work properly, we need to know the [DPI](https://en.wikipedia.org/wiki/Dots_per_inch) of the screen. Right now, it is hardcoded to 96:
https://github.com/plotly/orca/blob/1d56807e1146aa5a28d81ad46183b6f45346e96a/src/component/plotly-dash-preview/constants.js#L1-L2
This will not always be the case. I imagine that `xvfb` defaults to 96 so this is why this fix works in the Docker image. We should probably specify this DPI value in case the default changes in the future. It should also be noted that the current fix will not work for desktop users with a high-density display.

Finally, although I understand this needs to be fixed urgently, I am getting a bit uneasy about merging things that aren't tested on CI.

antoinerg:
> Actually, I think it is because we need to round off window size as well:

If that's the case, we should also round off the window size if the user specifies pageSize as a string: https://github.com/plotly/orca/blob/5c1f71d6d8ce046ec167997b722228820117f999/src/component/plotly-dash-preview/parse.js#L39-L40

antoinerg:
I don't think commit f2a2bad is necessary. It would be much cleaner and DRY to round off the numbers on line 59 just after the following block just before passing them to the renderer:
https://github.com/plotly/orca/blob/574241a3a9019c9c1580cf469c47857aaedd1ca6/src/component/plotly-dash-preview/parse.js#L39-L59

deployment/run_server
```diff
@@ -40,7 +40,7 @@ pkill -9 Xvfb
 pkill -9 node
 pkill -9 electron
 
-xvfb-run --auto-servernum --server-args '-screen 0 640x480x24' /var/www/image-exporter/bin/orca.js serve $REQUEST_LIMIT --safe-mode --verbose $PLOTLYJS_ARG $ORCA_IGNORECERTERRORS_ARG $@ &
+xvfb-run --auto-servernum --server-args '-screen 0 1024x768x24' /var/www/image-exporter/bin/orca.js serve $REQUEST_LIMIT --safe-mode --verbose $PLOTLYJS_ARG $ORCA_IGNORECERTERRORS_ARG $@ &
```
antoinerg: Is this really necessary? I thought this line: `createBrowserWindowOpts['enableLargerThanScreen'] = true` would make the result independent of the screen resolution :thinking:

tarzzz: I was under the same impression, but apparently it doesn't. Should I remove that option, or let it be there?

antoinerg: That is unfortunate... I wonder if that is fixed in recent versions of `electron` (see branch [`pr-bump-electron-6`](https://github.com/plotly/orca/tree/pr-bump-electron-6)). 

Anyway, changing the resolution of the screen seems like a short-term solution: what if a user needs a bigger browser size than 1024 by 768... We're essentially putting a limit on how big a page we can print. If so, it should be clearly stated somewhere. By the way, what do you mean when you say it "works" (ie. how/what are you testing exactly)?

antoinerg:
Okay so I can see that with your latest commit we go from a PDF with a figure not filling the page ([before.pdf](https://github.com/plotly/orca/files/3654822/before.pdf)) to one that does ([after.pdf](https://github.com/plotly/orca/files/3654823/after.pdf)).

antoinerg:
This is related to (another) issue: https://github.com/plotly/dash-snapshots/issues/62 . Looking at `parse.js`, it seems to me that if we pass `body.pageSize` as a string, we won't get the right result. It is a bit unfortunate in my opinion that we have two ways of specifying the size of the page (via `body.pageSize` and via `body.pdf_options.pageSize`) but we'll have to live with it.

Please test that for the 4 possible input body:
- `body.pageSize` is a string
- `body.pageSize` is an object with width and height
- `body.pdf_options.pageSize` is a string
- `body.pdf_options` is an object with width and height
we end up with a `result.browserSize` and a `result.pdfOptions.pageSize` that are equivalent (ie. same size). Finally, add a test to lock down the current behavior when both `body.pageSize` and `body.pdf_options.pageSize` are specified. Which one wins?

cc @tarzzz @ycaokris

tarzzz:
I have changed the base to `master` because we don't know yet if it needs to go to 3.3 release..

antoinerg:
@tarzzz By the way, I can confirm that this component mishandles 2 of the 4 scenarios described in https://github.com/plotly/orca/pull/250#issuecomment-535323595

antoinerg:
With commit https://github.com/plotly/orca/pull/250/commits/0943a790019855649f6d9d223395c603fd94f1e0 I introduced new tests that enabled me to discover bugs in the way we used to parge `pageSize`.

It fixes:
- [x] handling of `pageSize` defined as a string
- [x] handling of `pdf_options.pageSize` defined as an object with width and height in microns
- [x] It also guarantees that the browser size and the PDF page size are the same (assuming a DPI of 96) to ensure that no page reflow is necessary (essentially a workaround for issue https://github.com/plotly/plotly.js/issues/1275)

antoinerg:
@etpinard I'm taking over this PR so I might need you to review it.

antoinerg:
> I have changed the base to `master` because we don't know yet if it needs to go to 3.3 release.

If this doesn't go into 3.3, this means that snapshots will have figures that are not sized appropriately. I am pretty sure that @chriddyp and @ycaokris want this in. I will let them confirm whether this is the case or not.

cc @tarzzz

ycaokris:
> > I have changed the base to `master` because we don't know yet if it needs to go to 3.3 release.
> 
> If this doesn't go into 3.3, this means that snapshots will have figures that are not sized appropriately. I am pretty sure that @chriddyp and @ycaokris want this in. I will let them confirm whether this is the case or not.
> 
> cc @tarzzz

I'd like to know what's further steps we need to do in order to merge this into `3.3` release. If we're able to get it fixed before 3.3 due we could possibly skip temporary match for `3.3.2` https://github.com/plotly/dash-snapshots/pull/64

test/unit/plotly-dash-preview_test.js
```diff
@@ -2,9 +2,14 @@ const tap = require('tap')
 const sinon = require('sinon')
 
 const _module = require('../../src/component/plotly-dash-preview')
+const constants = require('../../src/component/plotly-dash-preview/constants')
 const remote = require('../../src/util/remote')
 const { createMockWindow } = require('../common')
 
+function clone (obj) {
+  return JSON.parse(JSON.stringify(obj))
+}
```
etpinard: I know this is only test code, but this `clone` function is dirty (and slow). 

It might not be worth adding another dev dep to use an "extend deep" helper, but here looks like you can simply turn `const mock` the object into a function returning a new mock object:

```js
const mock = () => ({
  url: 'https://dash-app.com',
  selector: 'dummy'
})
```

etpinard:
:dancer: thanks for taking care of this!

I made one non-blocking https://github.com/plotly/orca/pull/250#discussion_r329718833

test/unit/plotly-dash-preview_test.js
antoinerg: Done in 156a35cc5e6111a85910fa47fa6eb670db5708f8

PR #251, opened 2019-09-23
Dockerfile: add package-lock.json before npm install, copy Orca last :racehorse:

antoinerg:
It is much faster to copy Orca's source code last in order to hit the cache on all the first (rarely changing) build steps. This saves us roughly 8 minutes in the `docker-build-and-push` step on CircleCI now that we have Docker layer caching enabled.

Also, we copy `package-lock.json` prior to running `npm install` to get exactly the right dependencies in the Docker image.

cc @etpinard


etpinard:
Awesome :dancer:

PR #257, opened 2019-10-04
test-image: fix error reporting

antoinerg:
Right now on `master`, running test-image on CI will yield:
`grep: build/test_images_diff/*.txt: No such file or directory`. This regression was introduced in 697068a13a470c9b74d644f255fb2f72eeeddeed . The current PR fixes this issue.

In a nutshell, grep doesn't expand glob to match files but the shell will do it if it's not between double-quotes. Note that the script still passes a `shellcheck`.

### DIFF
--- a/test/image/compare_images
+++ b/test/image/compare_images
@@ -44,7 +44,7 @@ do
   cat "$3/$filename.txt"
 done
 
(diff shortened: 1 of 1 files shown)


etpinard:
:dancer: - love it - orca getting some maintenance :pick:

PR #262, opened 2019-10-21
Add CI config to automate onprem builds

tarzzz:
Part of https://github.com/plotly/streambed/issues/13214


antoinerg:
Screenshot_2019-10-21_10-51-30

tarzzz:
@antoinerg the error is because we had not added `GITHUB_TOKEN` in CI environment variable.
The variable has been added now so this should be good. 

@joaoalf Please review..

Issue #89, opened 2018-05-03
Try out Electron 2.0


etpinard:
I don't see anything breaking changes from the list on https://electronjs.org/releases#2.0.0, so the upgrade should be fairly simple. 

We might want to explore new possible solutions for https://github.com/plotly/orca/issues/12 and try out `--headless` as mentioned in https://github.com/plotly/orca/issues/22.

Some thorough testing of our image server will be required though.

etpinard:
... to do after the v1 release.

etpinard:
... and electron 3.0 and 4.0

- https://electronjs.org/blog/electron-3-0
- https://electronjs.org/blog/electron-4-0

PR #266, opened 2019-11-12
bump electron to 6.1.4

### LINKED ISSUE #124: Segmentation fault when NODE_OPTIONS environment variable set

### LINKED ISSUE #248: EPS: should preserve graphics in vector format and not rasterize
At the moment, opening EPS files produced by Orca in either Illustrator or Inkscape show raster graphics rather than a vector graphics. Zooming in shows that it is pixelated:
image (5)

antoinerg:
This PR supersedes PR https://github.com/plotly/orca/pull/237 and bumps electron from `1.8.4` to `6.1.4`. The version of the bundled Chromium goes from `v59` to `v76`.

- It fixes https://github.com/plotly/orca/issues/248 as can be seen by opening the new EPS baselines.
- It fixes #124 and makes PR https://github.com/plotly/orca/pull/125 unnecessary.
- It closes https://github.com/plotly/orca/issues/89

Although most SVG baselines stayed the same, there are slight differences in rendering for the other baselines (most certainly due to switching to a much newer version of Chromium!).

TODO:
- [x] Test that this PR closes #124 (it does)
- [x] Make sure the PDF output remains high quality for DDK. Please inspect the [following PDFs](https://circleci.com/gh/plotly/orca/4010#artifacts/containers/0) versus the [PDFs built from `master`](https://circleci.com/gh/plotly/orca/3933#artifacts/containers/0). cc @wbrgss 
- [x] remove modebars appearing on some of the figures (see https://github.com/plotly/orca/pull/266#issuecomment-566762220) (fixed by https://github.com/plotly/orca/commit/22032a431d154b60e352c46c417f8e0d4929361a)
- [x] Make sure we test the new Docker image on stage. After deploying on stage, I could test that exporting graphs to EPS images do not create a rasterized figure.

Nice to have:
- [ ] To completely close #248, we need to remove all semi-transparency in the original SVG (see https://github.com/plotly/orca/issues/248#issuecomment-533705025). This might potentially be implemented in https://github.com/plotly/plotly.js/issues/4433.


antoinerg:
Comparing PDF of DDK apps:
New: https://3931-99037241-gh.circle-artifacts.com/0/test_dash_previews/report_graph.pdf
Old: https://3933-99037241-gh.circle-artifacts.com/0/test_dash_previews/report_graph.pdf

- We can see that the white boxes now have a shadow. Is that intentional?
- Somehow, the first plot has the mode bar displayed :\ It seems to be case for other PDFs as well 
 such as [this one](https://3931-99037241-gh.circle-artifacts.com/0/test_dash_previews/report_graph_dimensions.pdf). This didn't seem to be the case...

cc @wbrgss @chriddyp

wbrgss:
@antoinerg The shadow should be there; it is on the web report. I'd say this is an improvement from the upgrade

The modebar shouldn't be there. I notice it's only on the first graph. Is is being focused or even "hovered" somehow?

antoinerg:
> The modebar shouldn't be there. I notice it's only on the first graph. Is is being focused or even "hovered" somehow?

This one is a strange regression. I'm surprised by how reproducible it is: the modebar always appears if a graph is near the top of the first page. I saw this issue before when working on `dash-snapshots` but it was intermittent. Anyway, it seems like https://github.com/plotly/orca/pull/266/commits/22032a431d154b60e352c46c417f8e0d4929361a fixes the issue :tada:

antoinerg:
Ok, so it seems like the major blocking issues have been resolved for this one and it should get merged in the near future.

@plotly/devops, I deployed this branch on stage and made sure it works in Chart studio. You might want to battle test it further to see how well it behaves under load.

etpinard:
:dancer:

etpinard:
_An early candidate for PR of the year 2020_

Issue #185, opened 2019-01-10
chromium vs google-chrome-stable


TiemenSch:
I noticed that the install instructions include quite an elaborate method of installing ``google-chrome-stable`` by adding Google's repository (and keys) before installing. 

Wouldn't it already be sufficient to just install ``chromium-browser`` from the Linux distro's default package manager?

I even experimented with installing the bare dependencies orca named in it's error messages, but suspect that could become problematic down the road.

antoinerg:
> I noticed that the install instructions include quite an elaborate method of installing google-chrome-stable by adding Google's repository (and keys) before installing.

Are you referring to this instruction: https://github.com/plotly/orca/blob/master/README.md#linux-troubleshooting-cannot-open-shared-object ? 

> Wouldn't it already be sufficient to just install chromium-browser from the Linux distro's default package manager?

I think you are right that `chromium-browser` is sufficient and probably even more appropriate since that's what Electron uses.

Nikolai-Hlubek:
+1 for updating the readme. 

The way IT departments in coorporations usually work is, they give you a virtual server that has access to an offline copy of your linux distribution but no further internet access.

So for me in this scenario using CentOS
> sudo yum install chromedriver chromium xorg-x11-server-Xvfb
worked. While downloading google-chrome-stable would have been some extra task.

PR #267, opened 2019-11-13
depend on chromium instead of google-chrome

antoinerg:
This PR closes https://github.com/plotly/orca/issues/185 by removing our dependency to `google-chrome-stable`. We can instead install `chromium-browser` which brings in all the necessary dependencies for `electron` to run properly.

Because `google-chrome-stable` also installs a special font package `fonts-liberation`, so I had to add it to the Dockerfile in order for figures to remain pixel perfect.


etpinard:
Brilliant :dancer:

PR #271, opened 2019-12-04
Recommend `xvfb-run -d` instead of `xvfb-run -a`

### LINKED ISSUE #270: Don't recommend deprecated `xvfb-run -a`
`-a` is deprecated:

```
-a        --auto-servernum          try to get a free server number, starting at
                                    --server-num (deprecated, use --auto-display
                                    instead)
-d        --auto-display            use the X server to find a display number
                                    automatically
```

and can cause problems. See https://github.com/plotly/orca/issues/239#issuecomment-561553654_

micheljung:
Fixes #270

### DIFF
--- a/README.md
+++ b/README.md
@@ -167,7 +167,7 @@ script that runs the orca AppImage executable using the `xvfb-run` command.
 
 ```
 #!/bin/bash
-xvfb-run -a /path/to/orca-X.Y.Z-x86_64.AppImage "$@"
+xvfb-run -d /path/to/orca-X.Y.Z-x86_64.AppImage "$@"
 ```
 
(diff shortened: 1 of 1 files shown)


antoinerg:
Thank you @micheljung for submitting a PR :smile_cat: 

In light of https://github.com/plotly/orca/issues/270#issuecomment-561815224, I think we should keep mention of the `-a` flag for older versions of `xvfb-run` (which are still prevalent). However, mentioning users should use `-d` for newer versions of `xvfb-run` is a great addition.

antoinerg:
@micheljung do you mind me taking over this PR and make the appropriate modifications?

micheljung:
@antoinerg not at all, go ahead :)

Issue #140, opened 2018-10-20
Wrong file permissions for orca-build in Linux


Ejdamm:
Hi,
We are several users using orca on the same system and we have to erase the orca-build folder in /tmp everytime a new user wants to run orca. This is because it is created with 755 as file permissions so other users can not write to it. 

The line that creates the folder is located in /src/util/create-index.js:10
Documentation for mkdirsync says default permission should be 777 so it should work as it is now, but it doesn't for me. Can someone test if this is a local problem for me or everyone? I'm thinking it might be something with umask but my knowledge about that area is limited. I don't know anything about Node.js either but I would try to override the default 777 with 755 and see it that changes anything. Note that the mode option is not available on Windows. If the mode is set, would windows complain or just ignore it?

Docs on Node.js mkdirsync: https://nodejs.org/api/fs.html#fs_fs_mkdirsync_path_options

jonmmease:
Cross reference forum discussion https://community.plot.ly/t/plotly-orca-connectionrefusederror/19723

I can confirm that on my ubuntu system the permission of the `/tmp/orca-build` directory are set to `drwxrwxr-x`, which I expect would cause a problem if there were another user on the system.

jonmmease:
Modification of `src/util/create-index.js` that worked for [`claudia`](https://community.plot.ly/u/claudia)

```javascript
try {
  fs.mkdirSync(PATH_TO_BUILD)
  fs.chmodSync(PATH_TO_BUILD, 0777)
} catch (e) {}
try{
  fs.readdir(PATH_TO_BUILD, (err, files) => {
  if (err) throw err

  for (const file of files) {
    fs.unlink(path.join(PATH_TO_BUILD, file), err => {
      if (err) throw err
    })
  }
})
} catch (e) {}
```

In addition to adjusting the permissions, they are explicitly deleting the temp files.  If orca is not already cleaning up these files on exit, then we should take a look at that also.

kevmk04:
I can also confirm that the solution by [claudia](https://community.plot.ly/u/claudia) also worked for me.  I'm using this on a TLJH installation with multiple users successfully.

LukaPitamic:
Hi guys, I still have problem on TLJH.
[...]

PR #310, opened 2020-04-09
Fix build directory permissions

justin-sleep:
Per discussion in #140, this change ensures that when orca's build directory is created, it is world-writeable. This is necessary on multi-user systems so that if one user creates the build directory, other users can use it.

### DIFF
--- a/src/util/create-index.js
+++ b/src/util/create-index.js
@@ -7,7 +7,7 @@ const os = require('os')
 const COMPONENT_GLOBAL = 'PlotlyExporterComponent'
 const PATH_TO_BUILD = path.join(os.tmpdir(), 'orca-build')
 try {
-  fs.mkdirSync(PATH_TO_BUILD)
+  fs.mkdirSync(PATH_TO_BUILD, 0o777)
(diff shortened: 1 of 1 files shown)


src/util/create-index.js
```diff
@@ -7,7 +7,7 @@ const os = require('os')
 const COMPONENT_GLOBAL = 'PlotlyExporterComponent'
 const PATH_TO_BUILD = path.join(os.tmpdir(), 'orca-build')
 try {
-  fs.mkdirSync(PATH_TO_BUILD)
+  fs.mkdirSync(PATH_TO_BUILD, 0o777)
```
antoinerg, line 10: @justin-sleep I'm a bit confused: isn't that supposed to be the default already?

justin-sleep: The default is 777, but then the default umask is applied:
```JavaScript
> process.umask().toString(8)
'22'
```
leaving the directory with permissions of 755.

antoinerg:
Thank you @justin-sleep !

PR #315, opened 2020-05-01
test-image: bump plotly.js to 1.54.0

antoinerg:
With the introduction of shape selection in plotly.js 1.54.0, there's a new default layout attribute called `activeshape`. Because of it, I needed to update the JSON baselines. There are no visual differences in either of our other image export format.

cc @alexcjohnson @archmoj


archmoj:
LGTM

Non-blocking: 
To make it easier the compare and keep track of the changes using git/gitHub is there an easy way to post-process the `json` export and write the file on different lines?
Something like:
```
JSON.stringify(rawObject, null, 1)
```

antoinerg:
I added `jq` to the container running the image tests and use it to prettify and sort JSON files prior to diffing. This leads to diff like this:
```
3999,4002d3998
<     "activeshape": {
<       "fillcolor": "rgb(255,0,255)",
<       "opacity": 0.5
<     },
4175,4186d4170
<     "newshape": {
<       "drawdirection": "diagonal",
<       "fillcolor": "rgba(0,0,0,0)",
<       "fillrule": "evenodd",
<       "layer": "above",
<       "line": {
<         "color": "#444",
<         "dash": "solid",
<         "width": 4
<       },
<       "opacity": 1
<     },
4335c4319
<   "version": "1.54.0"
---
>   "version": "1.53.0"
```

See https://github.com/plotly/orca/pull/315/commits/291de3aee3221a79b71b45c54b1eff35004d6139 and https://github.com/plotly/orca/pull/315/commits/d4bd4ed780bac3ec148812c8656c941f35d01da3

cc @archmoj @alexcjohnson

archmoj:
Thanks!
Nicely done.
--------------------------------------------------------------------------------------------------------------------:dancer:

PR #323, opened 2020-05-20
dash-preview: gracefully fail on pages with iframes

antoinerg:
Supersedes #316 

It seems that sometimes `printToPDF` does not complete and hangs forever (see https://github.com/electron/electron/issues/20634). This is problematic because the window stays open forever. If you get this problem often, there will eventually be too many windows opened and you will get an out of memory error.

This PR introduces a change that will automatically close the window after a timeout (either the one provided or a default one). The rationale is that printing to PDF should not take longer than the initial page load so we can rely on the user's provided timeout. 

A more elegant approach would be to perform cleanup at the component level whenever the HTTP server drops a client request because it's taking too long. At the moment, we keep working even if the connection is dropped...

cc @alexcjohnson

### DIFF
--- a/.circleci/test.sh
+++ b/.circleci/test.sh
@@ -7,7 +7,7 @@ set +e
 set +o pipefail
 
 EXIT_STATE=0
-MAX_AUTO_RETRY=5
+MAX_AUTO_RETRY=2
 
 # inspired by https://unix.stackexchange.com/a/82602
 retry () {
--- a/package-lock.json
+++ b/package-lock.json
@@ -20,23 +20,17 @@
       }
     },
     "@babel/generator": {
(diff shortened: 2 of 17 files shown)


alexcjohnson:
The title says "fix handling of pages with iframes" but the test just loads a page with iframes and tests that it fails gracefully. Are iframes just not supported?

antoinerg:
> The title says "fix handling of pages with iframes" but the test just loads a page with iframes and tests that it fails gracefully. Are iframes just not supported?

Sorry for the misleading title. Before this PR, a page with iframes would hang on forever and its window would never close. Eventually, this would lead to an out of memory error. This PR only add graceful failure.

PR #346, opened 2020-09-21
Support k8s >= 1.16

scjody:
Switch to the `apps/v1` API; all the parameters we use are still supported.

Part of https://github.com/plotly/streambed/issues/14300

* [x] Tested on stage
* [x] Won't affect On-Prem

1 - 3lotSp5


scjody:
@BRONSOLO Please review.

@antoinerg Image tests are failing due to some image differences that [you can see here](https://app.circleci.com/pipelines/github/plotly/orca/71/workflows/ea6bf5a0-2947-4a87-a732-654ce872f9a7/jobs/4594/artifacts). Is it OK to merge with these differences? Should I update the baselines? Something else?

BRONSOLO (approved):
:woman_dancing:

antoinerg:
> @antoinerg Image tests are failing due to some image differences that [you can see here](https://app.circleci.com/pipelines/github/plotly/orca/71/workflows/ea6bf5a0-2947-4a87-a732-654ce872f9a7/jobs/4594/artifacts). Is it OK to merge with these differences? Should I update the baselines? Something else?

There is only one slight difference in font rendering (the difference is in red below):
fonts pdf diff

I think it's OK to merge with these differences although I'm a bit puzzled as to what caused the difference. Anyway, I can update the baselines for the `fonts` mock in a subsequent PR. @scjody

